PR #16, opened 2015-07-20
Use Java 8 as default JDK instead of Java 7.

pmather:
Sufia 6.2.0 brought in an explict requirement for Java 8 for the
bundled hydra-jetty.  Fedora 4.2.0 also now requires Java 8.
It was decided we should move to using Java 8 as the default JDK.

This update uses Oracle Java 8 instead of OpenJDK 8, because
currently Ubuntu Trusty does not have any OpenJDK 8 packages
available in its main repositories or in backports.  (Utopic
Unicorn onwards have OpenJDK 8 packages, so we'll have to wait
until the next Ubuntu LTS release to avail ourselves of those
packages.  For now, we use the Oracle Java 8 packages via the
webupd8team PPA.)

We also make Java 8 the default JDK via alternatives.  Previously,
FITS installed Java 7.  According to the FITS documentation, FITS
should work with Java 6 and up.

### DIFF
--- a/Vagrant/bootstrap.sh
+++ b/Vagrant/bootstrap.sh
@@ -41,8 +41,15 @@ add-apt-repository -y ppa:brightbox/ruby-ng
 apt-get update
 apt-get install -y "$rubyver"
 
+# Install Java 8 and make it the default Java
+add-apt-repository -y ppa:webupd8team/java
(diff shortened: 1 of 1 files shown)


rotated8:
:+1:

PR #19, opened 2015-07-22
Remove old info, use the resque script, and don't start Webrick.

rotated8:


rotated8:
This should probably wait for #18 to be merged first.

PR #20, opened 2015-07-22
Do not port forward Webrick, but do forward Hydra Jetty

rotated8:


rotated8:
This probably should have been in #19

rotated8:
Closing, for @pmather's coming changes.

PR #21, opened 2015-07-23
Run application via Nginx/Passenger by default.

pmather:
This change moves the CAS_extras.sh steps into the main bootstrap.sh
install script.  This means the Sufia application will be run via
Nginx/Passenger rather than the WEBrick server used by "rails server."
The other significant change is that now all the installation steps are
automated: there is no need to run CAS_extras.sh manually after the
initial provisioning.

This commit includes these specific changes:
- The version of Sufia has been bumped to 6.2.0.
- The WEBrick port is no longer exposed to the host system in Vagrant.
- The Fedora/Solr port (8983) is exposed to the host system in Vagrant.
- The Web server port mappings are changed in Vagrant.  They are now
  8080 -> 80 and 4443 -> 443.  This is because the Web server now runs on
  standard ports on the VM/server.
- Because it has been incorporated into bootstrap.sh, the
  CAS_extras.sh script has been removed.
- The bootstrap.sh script now takes a fourth argument: the hostname of
  the Web server site (default: "localhost").
- Nginx and Passenger are installed via the Phusion Passenger APT
  repository as part of the main bootstrap sequence.
- Additionally, a Web site is configured that runs the Sufia
  application via Passenger.
- A self-signed SSL key is generated and installed for the
  aforementioned Web site, allowing HTTPS access.
- The Sufia application is deployed in production mode as part of the
  main bootstrap sequence instead of via CAS_extras.sh.
- Instructions are no longer issued to the user to generate a SSL
  certificate and restart Nginx.
- The "run always" Vagrant provisioner now also restarts Nginx, in
  case it is not running.


rotated8:
:+1:

PR #22, opened 2015-07-23
Add param for the application's environment variable.

rotated8:
Acceptable values are 'development' (the default), and 'production'.

### DIFF
--- a/Vagrant/Vagrantfile
+++ b/Vagrant/Vagrantfile
@@ -1,14 +1,13 @@
-# vi: set ft=ruby :
-
 installuser = "vagrant"
 hydrahead = "data-repo"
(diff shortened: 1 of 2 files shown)


rotated8:
cc @pmather: Does this seem like a sane approach to keeping Vagrant boxes as dev machines?

pmather:
I think it is worthwhile to try and keep the possibility of supporting both development and production Rails environments in Vagrant.  I believe your approach is as good as any.  I like the fact that it defaults to production, too.

pmather:
:+1: after commit is revised to incorporate suggestion.

rotated8:
Updated!

PR #23, opened 2015-07-31
Replace bundled hydra-jetty with standalone Solr and Fedora 4 services.

pmather:
This change alters the installation script so that hydra-jetty is no
longer installed and used.  Instead, a separate Solr and Fedora 4 are
installed and used by the application.

Currently, we install Solr 5.2.1 and Fedora 4.2.0.  Fedora is built
from a specific source branch, as per request.  The Solr installation
uses the install_solr_service.sh script that comes bundled with Solr 5.
Fedora is installed to be run via Tomcat 7.

Tomcat 7 runs on port 8080 and Solr on port 8983.  Solr installs under
/opt/solr and runs under user "solr".  Its cores and related server
runtime data are kept under /var/solr.  Fedora's data is stored under
$installuser/fedora-data.

Here are some specific changes:
- Hydra-jetty is no longer started explicitly in the Vagrantfile
  runserver script.
- Local port 8888 is forwarded to the Tomcat 7 service (port 8080) in
  the Vagrantfile.
- The major version number in bootstrap.sh has been bumped.
- The version of Rails that is installed has been locked to Rails
  4.2.x.
- Solr installs under its own user, as per the install_solr_service.sh
  script.  It was kept at the default "solr" so as better to isolate Solr
  from the application.
- Beware: Fedora 4 installs with the default fedoraAdmin password.
- Fedora 4 is deployed to Tomcat 7 as a WAR file built from source.
- Hydra-jetty is no longer installed, configured, and started.
- The path to the FITS executable is now set explicitly in
  config/initializers/sufia.rb.
- Solr cores are created for the application.  The core name
  corresponds to the Rails environment under which the application is
  running, i.e., development or production.
- Application configuration files config/solr.yml,
  config/blacklight.yml, and config/fedora.yml are patched to point to
  the standalone Solr and Fedora services.
- Solr and Tomcat services are stopped after install whilst the
  application is being installed and configured.  Their services are
  started after everything is installed.

### DIFF
--- a/Vagrant/bootstrap.sh
+++ b/Vagrant/bootstrap.sh
@@ -3,7 +3,7 @@ set -x -o errexit
 
 # For Ubuntu Server 14_04
 # Installs the default Sufia application and all of it's dependencies.
-echo "Version 2.0"
+echo "Version 3.0"
 
 # Vars
 installuser="ubuntu" # Name of user to install under (must already exist)
@@ -38,13 +38,30 @@ SSL_CERT_DIR="/etc/ssl/local/certs"
 SSL_CERT="$SSL_CERT_DIR/$hydrahead-crt.pem"
 SSL_KEY_DIR="/etc/ssl/local/private"
 SSL_KEY="$SSL_KEY_DIR/$hydrahead-key.pem"
+SOLR_USER="solr" # User under which Solr runs.  We adopt the default, "solr"
+SOLR_VERSION="5.2.1"
+SOLR_MIRROR="http://www.gtlib.gatech.edu/pub/apache/lucene/solr/$SOLR_VERSION/"
+SOLR_DIST="solr-$SOLR_VERSION"
+SOLR_INSTALL="/opt"
+SOLR_MUTABLE="/var/solr"
+SOLR_DATA="$SOLR_MUTABLE/data" # Where Solr cores live
+TOMCAT_CONF="/etc/tomcat7"
+FEDORA4_REPO="https://github.com/fcrepo4/fcrepo4.git"
+FEDORA4_BRANCH="fcrepo-4.2.0"
+FEDORA4_VER="4.2.0"
+FCREPO4_DATA="$basedir/fedora-data"
+FCREPO4_APP_DIR="/var/lib/tomcat7/webapps"
+FCREPO4_USER="tomcat7"
+FCREPO4_GROUP="tomcat7"
+JDK_HOME="/usr/lib/jvm/java-8-oracle"
 
 fitsver="fits-0.6.2" # The version of FITS to install.
 rubyver="ruby2.2" # The version of Ruby to install.
-railsver=">=4.2" # The version of Rails to install.
+railsver="~> 4.2" # The version of Rails to install.
 sufiaver="6.2.0" # The version of Sufia to install.
 
 RUN_AS_INSTALLUSER="sudo -H -u $installuser"
+RUN_AS_SOLR_USER="sudo -H -u $SOLR_USER"
 
 # Update packages
 cd "$basedir/"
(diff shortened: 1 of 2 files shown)


pmather:
You should probably drop a "2.0" tag (or "hydra-jetty" or whatever seems an appropriate name) on the master branch before merging this pull request.  That way, people will be able to pull easily the old installation method that uses the bundled hydra-jetty.

Vagrant/bootstrap.sh
```diff
@@ -85,6 +102,53 @@ apt-get install -y nodejs
...
+if ! grep -q 'fcrepo.home=' /etc/default/tomcat7 ; then
+  echo "JAVA_OPTS=\"${JAVA_OPTS} -Dfcrepo.home=$FCREPO4_DATA\"" >> /etc/default/tomcat7
+fi
+# Create Fedora data directory and make sure Tomcat 7 can write to it
+mkdir -p $FCREPO4_DATA
+chown ${FCREPO4_USER}:${FCREPO4_GROUP} $FCREPO4_DATA
+chmod 770 $FCREPO4_DATA
+# Build Fedora 4 from sources
+apt-get install -y git maven
+$RUN_AS_INSTALLUSER git clone --branch $FEDORA4_BRANCH --depth 1 $FEDORA4_REPO
```
rotated8, line 143: Since the repo is a configuration variable, this should probably specify a folder to clone into, so the `cd` and `rm` below always know what to operate on.

Vagrant/bootstrap.sh
```diff
@@ -215,7 +276,29 @@ if [ "$app_env" == "production" ]; then
     $RUN_AS_INSTALLUSER RAILS_ENV=production bundle exec rake assets:precompile
     $RUN_AS_INSTALLUSER RAILS_ENV=production bundle exec rake datarepo:setup_defaults
 fi
+# Fix up configuration files
+# 1. FITS
+$RUN_AS_INSTALLUSER sed -i "s@config.fits_path = \".*\"@config.fits_path = \"$fitsdir/$fitsver/fits.sh\"@" config/initializers/sufia.rb
```
rotated8, line 281: Was this causing you trouble?

Vagrant/bootstrap.sh
```diff
@@ -215,7 +276,29 @@ if [ "$app_env" == "production" ]; then
...
+# 4. Make the blacklight.yml file point to an appropriate $app_env core
+sed -i '/production:/ {N; N; s@^production:\(.*\)/development@production:\1/production@}' config/blacklight.yml
+# 5. Make the fedora.yml point to Tomcat 7 port, not to hydra-jetty port 8983
+sed -i 's/url:\(.*\):8983/url:\1:8080/' config/fedora.yml
 $RUN_AS_INSTALLUSER bash "$hydradir/scripts/restart_resque.sh" "$app_env"
-$RUN_AS_INSTALLUSER bundle exec rake jetty:start
-# Start Nginx
+# Start services
+service tomcat7 start
+service solr start
```
rotated8, line 303: Should tomcat and solr be restarted in the Vagrantfile as well?

rotated8:
I'll add a 2.0.0 release once this is merged. Just a couple things to look at first. Thanks Paul!

rotated8:
Is the Fedora compilation the one downloading everything it can get from maven?

Vagrant/bootstrap.sh
pmather, line 143: Good observation.  The Fedora 4 install instructions explicitly refer to the fcrepo4 directory, and so that's where I think that came from.  But, if, ultimately, we build from our own Fedora 4 repository, we might end up renaming it, so your improvement would be necessary.

Vagrant/bootstrap.sh
pmather, line 281: It wasn't currently causing trouble, but hard-coding it to /home/vagrant/... in the sufia.rb file means it won't work in AWS or in the case where $fitsdir points to somewhere other than /home/vagrant...

This change ensures the value in the sufia.rb config file matches the location used in the installation script.

Vagrant/bootstrap.sh
pmather, line 303: I left that out because, as we've discussed offline, arguably the Nginx shouldn't be restarted in the Vagrantfile runserver script.  All three services start at system boot, so restarting them there is redundant.

pmather:
Building Fedora from source causes LOTS of files to be downloaded.  That's probably why most people just download a pre-built Fedora WAR file. :smiley:

Vagrant/bootstrap.sh
```diff
@@ -215,7 +276,29 @@ if [ "$app_env" == "production" ]; then
...
+# Make links to keep the Hydra Solr solrconfig.xml paths happy
+$RUN_AS_SOLR_USER ln -s $SOLR_INSTALL/solr/contrib
+$RUN_AS_SOLR_USER ln -s $SOLR_INSTALL/solr/dist
+$RUN_AS_SOLR_USER mkdir lib
+$RUN_AS_SOLR_USER ln -s $SOLR_INSTALL/solr/contrib lib/contrib
+cd $hydradir
+# 3. Make the solr.yml file point to an appropriate $app_env core
+sed -i '/production:/ {N; s@^production:.*development@production:\n  url: http://localhost:8983/solr/production@}' config/solr.yml
+# 4. Make the blacklight.yml file point to an appropriate $app_env core
+sed -i '/production:/ {N; N; s@^production:\(.*\)/development@production:\1/production@}' config/blacklight.yml
```
rotated8, line 297: This hurt my head looking at it the first time, I think I understand it now. It's changing the production urls to point to the production cores, right?

rotated8:
:ok:

PR #24, opened 2015-08-12
Major refactoring and support installation on AWS.

pmather:
This commit is a major refactoring of the existing Vagrant-only
bootstrap.sh script so installation can be supported on AWS.  Part of
the refactoring is to extract many configurable settings out into
"configuration files" and to support Vagrant and AWS settings that can
override common defaults.

The old bootstrap.sh script is now a higher-level driver for the
Vagrant and AWS installs.  The previous bootstrap.sh now becomes
bootstrap_server.sh, and has itself been refactored to separate out
installation of the major components such as Solr and Fedora 4 into
separate installation scripts.  This makes a clearer demarcation as to
what is happening in those component installations, and also preps for
the time when we will need to install them on separate servers.

A "files/" directory has been added to put local configuration
settings/files that override defaults.  The main intent for this
directory is for assets that shouldn't be committed to the public
repository such as Web server certificates/keys and ORCID API secrets.

Deployment is done via provisioning scripts and associated files.
Vagrant makes it easy to pass these files to the server being
provisioned via a shared folder (i.e., "/vagrant").  AWS allows
provisioning material to be passed via "userdata", which is limited to
about 16 KB.  As such, for AWS, the necessary files are packed into a
self-extracting script which is passed to the server being brought up
in AWS.  This script unpacks itself and then runs bootstrap_server.sh
to install the application and its prerequisites.  AWS appears to
reject binary data in "userdata" scripts, which is why the payload data
for the self-extracting "userdata" script is base64-encoded.  (This is
normally not necessary, but is required by awscli.)

In the newly-refactored scripts, the server provisioning scripts all
expect to be passed two arguments: the first is the target platform
under which installation is being performed ("aws" or "vagrant"); the
second is the path to the directory in which the install scripts live.
The bootstrap.sh script, that the user runs to install either on
Vagrant or AWS, takes only a single argument: "aws" or "vagrant".

This commit also includes an initial README giving an overview of the
install scripts and how to use them.

In the refactoring, the original UbuntuServer_1404_Sufia_Install.sh
script has been removed, as it is not longer relevant to the current
state of the application.  Also, the "Vagrant" directory has been
removed and all relevant files moved into the main InstallScripts
directory.


bootstrap.sh
```diff
@@ -0,0 +1,84 @@
...
+  aws ec2 associate-address --instance-id $ID --allocation-id $AWS_ELASTIC_IP
+  # Set descriptive name tag
+  aws ec2 create-tags --resources $ID --tags Key=Name,Value=$SERVER_HOSTNAME
+  # Clean up after ourselves
+  rm -rf $TMPDIR
+}
+
+# Process script arguments. Argument is server type: vagrant or aws; default: aws
+SCRIPTS_DIR="`pwd`"
+SERVER_ENV="vagrant"
```
rotated8, line 54: This seems to contradict the comment above.

pmather, line 54: Good catch.  The comment is wrong.  Do you want me to submit a new pull request (re-typing in the humongous commit message :weary: ) or can I promise to fix it in the next update? :smiley:

rotated8:
:+1:

PR #25, opened 2015-08-13
Allow EBS root volume size to be set on instance launch.

pmather:
Add an AWS_EBS_SIZE configuration variable to specify the size in GB of
the EBS root volume of the server being provisioned in AWS.


rotated8:
:+1:

PR #26, opened 2015-08-13
Allow larger uploads in Nginx

pmather:
Configure Nginx to allow uploads up to 200 MB via the
client_max_body_size directive.


rotated8:
:+1:

PR #28, opened 2015-08-20
Use pre-built Fedora 4 WAR instead of building from source

pmather:
Instead of building Fedora 4 from source during the installation, we
now fetch a pre-built WAR file from the URL given in FEDORA4_WAR_URL.
This not only speeds up the installation but also makes it more robust
to transient built failures.

For now, we are using the official Fedora 4 release distribution WAR.
In future, this URL will point to our own Fedora 4 WAR built separately.


rotated8:
:+1:

PR #31, opened 2015-08-20
Enable sending e-mail via Postfix

pmather:
Install Postfix and configure for sending e-mail.  Opportunistic
encryption is enabled for sending e-mail via "smtp_tls_security_level =
may" option.


rotated8:
:+1:

PR #32, opened 2015-09-01
Enforce some resource limits on Nginx/Passenger

pmather:
We do some tuning of the Nginx/Passenger resource consumption in order
to limit the amount of memory used overall.  Also, we pre-start our
application to give faster startup times.

These are the specific tunings added in this commit:
- Set the number of Nginx workers to "auto" so as to match the number
  of cores in the host system;
- Start up exactly $PASSENGER_INSTANCES Passenger workers by means of
  the passenger_max_pool_size and passenger_min_instances directives.
  Normally, we would want to set $PASSENGER_INSTANCES to the number of
  cores in our target system;
- Set up a requests limiter zone such that each source IP address is
  limited to $NGINX_CLIENT_RATE requests.  Also, $NGINX_CLIENT_BURST can
  be used to specify a burstiness to the average rate limit, such that an
  individual source IP can exceed the average rate by this burst amount
  without delay and without provoking a 503 return code.


rotated8:
:+1:

PR #34, opened 2015-09-02
Adjust Solr logging settings

pmather:
Adjust the logging settings in /var/solr/log4j.properties so they are
more like those in the "Taking Solr to Production" guide
(https://cwiki.apache.org/confluence/display/solr/Taking+Solr+to+Product
ion#TakingSolrtoProduction-Logsettings)

Specifically, we do the following:
- Disable the CONSOLE logger;
- Set the log rotation size threshold to a larger value
  ($SOLR_LOGSIZE).

(We also remove some redundant config.sh settings which should be set
in config_PLATFORM.sh, namely APP_ENV and SOLR_CORE.)


rotated8:
:+1:

PR #35, opened 2015-09-03
Tune request rate limiting defaults

pmather:
The former defaults for Nginx limit_req were far too conservative for
the application working in RAILS_ENV "development" mode.  This resulted
in a lot of 503 errors and mis-rendering of the application pages until
the client browser could cache application assets locally (particularly
Javascript).

Here, we set a large default burst size so that typical clients can go
to a page in the application without trouble.  In RAILS_ENV
"production" settings, the burst size and NGINX_CLIENT_RATE can likely
be lowered.

Also, we allow the option to enable or omit "nodelay" on the limit_req
directive itself.


rotated8:
:100:

PR #36, opened 2015-09-04
Set Google Analytics ID

pmather:
Set the Google Analytics ID in config/initializers/sufia.rb if a
file/google_analytics_id file is present and we are not installing via
Vagrant.  The ID itself should be the only thing in the
google_analytics_id file.

Note, we uncomment the config.google_analytics_id setting and set it in
two steps in case the configuration setting has already been
uncommented from the default.  We can then set it to our chosen value.


rotated8:
:+1:

PR #38, opened 2015-09-25
Use PostgreSQL as the database adapter

pmather:
Install PostgreSQL and use it as the database adapter.  This required
some change in the order in which things are installed.  As part of
that, the installation of gems was modified after scrutinising the
existing approach.  Now there is only a single "bundle install".

Note that the existing install_postgresql.sh approach is to assume a
PostgreSQL setup where connections are made locally via a Unix domain
socket for performance.

Also, install_postgresql.sh is called from install_sufia_application.sh
because the PostgreSQL installation touches the application file
database.yml.


pmather:
This needs VTUL/data-repo#91 to be merged first.

PR #40, opened 2015-10-14
Fix install_clamav.sh permissions

pmather:
The install_clamav.sh committed in pull request #39 was accidentally
not executable.  This change fixes the permissions.


rotated8:
:+1:

PR #42, opened 2016-01-12
Use new rake tasks.

rotated8:
Remove old default setup task. Add new tasks to setup default roles,
populate users, and upgrade users to admins.

To create new users, add their email addresses (one per line) to a file
called 'user_list.txt' in the files directory. For administrators, the
file is called 'admin_list.txt', and works the same.


rotated8:
Depends on VTUL/data-repo#108

PR #50, opened 2016-02-08
Specify Google Analytics ID in secrets.yml

pmather:
This remaining fixup code is unnecessary after applying the
corresponding change to data-repo that sets config.google_analytics_id
if it is defined as a setting in config/secrets.yml


rotated8:
:+1: when vtul/data-repo#115 passes.

PR #58, opened 2016-02-25
Add test Solr core for development.

rotated8:
This allows developers to run the tests without having to create the
test core themselves.

### DIFF
--- a/config_aws.sh
+++ b/config_aws.sh
@@ -10,6 +10,7 @@ APP_ENV="production"
 #AWS_ELASTIC_IP="eipalloc-4fc4932a" # 54.209.80.216 - datadev.lib.vt.edu
 #APP_ENV="development"
 SOLR_CORE="$APP_ENV"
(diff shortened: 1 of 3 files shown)


config_aws.sh
```diff
@@ -10,6 +10,7 @@ APP_ENV="production"
 #AWS_ELASTIC_IP="eipalloc-4fc4932a" # 54.209.80.216 - datadev.lib.vt.edu
 #APP_ENV="development"
 SOLR_CORE="$APP_ENV"
+TEST_CORE="test"
```
pmather: If it wouldn't be too much trouble, could we change TEST_CORE to SOLR_TEST_CORE?  That would maintain the current convention of naming all the Solr-related settings SOLR_*.

If you don't want to, and want to keep it as TEST_CORE, let me know and I can merge as-is.

rotated8:
Updated.

pmather:
:+1:

Issue #8, opened 2015-02-12
Don't use curl to immadiately pipe to bash.


rotated8:
Replace the nodesource setup script with it's components.

rotated8:
Piping directly from curl to bash is not secure.

PR #67, opened 2016-07-14
Update dependencies

rotated8:


rotated8:
- Updates Fedora to 4.5.1
- Updates NodeJS to 6.x
- Updates Ruby to 2.3
- Adds PhantomJS caching for 2.1.1
- Removes Node's old curl-pipe-to-bash install, for a manual approach.

To use the PhantomJS caching, download the tar.bz2 for linux from their website, and put it into the ./files directory.

rotated8:
Fixes #8

rotated8:
Also, a gem handles installing PhantomJS if it is not cached, so it is no longer installed by apt-get.

pmather:
Verified OK. 👍

PR #70, opened 2016-07-22
Adds a synced folder for virtualbox

rotated8:
- Syncs ../data-repo/ on the host to /home/vagrant/data-repo on the VM.
- ../data-repo will be emptied when the VM is provisioned.
- The synced folder is only created for the virtualbox provider.
- The folder can be disabled by changing the disabled setting to true in the Vagrantfile.
- Sendfile is disabled for nginx, to prevent file corruption in the synced folder.
- Using this feature requires rebuilding the VM.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -81,6 +81,22 @@ fi
 [ -f "${SCRIPTS_DIR}/config.sh" ] && . "${SCRIPTS_DIR}/config.sh"
 [ -f "${SCRIPTS_DIR}/config_${SERVER_ENV}.sh" ] && . "${SCRIPTS_DIR}/config_${SERVER_ENV}.sh"
 
+# Check for sync folder, and make sure it's empty
+if [[ -d "../${HYDRA_HEAD}/" && $(ls -A "../${HYDRA_HEAD}/") ]]; then
+    echo "The local directory '../${HYDRA_HEAD}/' exists, and is not empty."
+    echo "If you continue, this directory will be deleted. You may lose your work."
+    # Only accept an answer for five seconds.
+    read -t 5 -p 'Do you want to continue? [y|N] ' REPLY
+    # Only proceed if something was read, and it was some form of yes.
+    if [[ $? -eq 0 && $REPLY =~ ^[Yy]([Ee][Ss])?$ ]]; then
+        echo "Deleting ../${HYDRA_HEAD}"
+        rm -rf "../${HYDRA_HEAD}"
+    else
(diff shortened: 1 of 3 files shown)


rotated8:
Removing everything from a synced folder is not the most subtle approach. This needs discussion.

rotated8:
This has been updated to confirm the local folder deletion.

rotated8:
Positive confirmations include 'Y<enter>', 'yes<enter>', or 'yEs<enter>'
Negative confirmations do not hit enter, or do not have one or three characters.

bootstrap.sh
```diff
@@ -81,6 +81,22 @@ fi
 [ -f "${SCRIPTS_DIR}/config.sh" ] && . "${SCRIPTS_DIR}/config.sh"
 [ -f "${SCRIPTS_DIR}/config_${SERVER_ENV}.sh" ] && . "${SCRIPTS_DIR}/config_${SERVER_ENV}.sh"
 
+# Check for sync folder, and make sure it's empty
+if [[ -d "../${HYDRA_HEAD}/" && $(ls -A "../${HYDRA_HEAD}/") ]]; then
+    echo "The local directory '../${HYDRA_HEAD}/' exists, and is not empty."
+    echo "If you continue, this directory will be deleted. You may lose your work."
+    # Only accept an answer for five seconds.
+    read -t 5 -p 'Do you want to continue? [y|N] '
```
pmather: Does this need a REPLY added to the end?

bootstrap.sh
```diff
@@ -81,6 +81,22 @@ fi
...
+    # Only proceed if something was read, and it was some form of yes.
+    if [ $? -eq 0 && $REPLY =~ ^[Yy]([Ee][Ss])?$ ]]; then
```
pmather: I am getting this error when testing:

./bootstrap.sh: line 91: syntax error near unexpected token `('

So, either REPLY is not being set or else the regex needs to be quoted?

pmather: I discovered this needs to be "if [[ ..." not "if [ ...".  The rest is okay.

rotated8: Oh, huh. I don't know how that happened...

bootstrap.sh
rotated8: REPLY is the default. I can add it to be explicit.

rotated8:
Updated: fixed Paul's errors.

pmather:
Verified OK. 👍

PR #71, opened 2016-08-11
Changes the provisioner to Ansible.

rotated8:


rotated8:
What do you think, @kayiwa ?

kayiwa:
👍

PR #79, opened 2016-08-19
Uses fetch for caching files.

rotated8:
I prefer to put the cached versions in their role's files dir because it keeps the file close to the role responsible for it (and role_path is a convenient variable), but am open to other options.


pmather:
I tend to agree with you when it comes to cached downloads like the PhantomJS distribution.  However, I'd rather not put something like a certificate file/key in the role directory but would instead rather have it separated out into a central `local_files/` or `secret_files/` directory as something that is supposed to be supplied externally, as was the semantics of the original bash scripts.

pmather:
Merging anyway, despite the above reservation.

PR #82, opened 2016-08-22
Incorporate GeoBlacklight support into Ansible playbooks

pmather:
This adds support for installing the GeoBlacklight project via the
Ansible playbooks.

GeoBlacklight support adds a number of roles, including: geoblacklight;
chroot-sftp; and tls_cert.

The Vagrantfile has been modified to support the selection of which
application type to install and which provisioner to be used.  The
defaults are "sufia" and "ansible_local" respectively.

The ANSIBLE_PROVISIONER environment variable may be set prior to
"vagrant up" to select which version of the Vagrant Ansible provisioner
is used: ansible or ansible_local.

If the APP_TYPE environment variable is set prior to "vagrant up" then
it is used to determine which type of application to provision if it is
set to a known type: "sufia" or "geoblacklight".  If it is not set, or
is not set to a valid application type then the value used is
determined by examining the "project_name" setting in the
ansible/site_secrets.yml file.


ansible/roles/common/tasks/main.yml
```diff
@@ -5,6 +5,11 @@
     state: present
     update_cache: yes
   with_items:
+    - acl
     - build-essential
     - apt-transport-https
     - software-properties-common
+    - git
```
rotated8: These four at the bottom should be in the GeoBlacklight role.

ansible/roles/tls_cert/README.md
```diff
@@ -0,0 +1,28 @@
+Tls_cert
```
rotated8, line 1: This role will get some changes when the duplicate part in the Sufia role is removed.

rotated8:
I rebased off master, and that removed the merge commit.

PR #84, opened 2016-08-24
Set GeoBlacklight application secrets from Ansible site_secrets.yml

pmather:
This allows settings in config/secrets.yml to be set based up project
settings in the Ansible site_secrets.yml file.  Previously, only
"secret_key_base" was set (randomly).  This commit adds more
configurable settings to config/secrets.yml and sets them with actual
site_secrets.yml configuration data.

### DIFF
--- a/ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
+++ b/ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
@@ -16,19 +16,13 @@
     force: yes
   become_user: "{{ project_user }}"
 
-- name: check for secrets file
-  stat:
-    path: "{{ project_app_root }}/config/secrets.yml"
-  register: secret
-
-- name: openssl key
-  command: /usr/bin/openssl rand -hex 64
-  register: ssl_result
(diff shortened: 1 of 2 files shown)


ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
```diff
@@ -16,19 +16,13 @@
     force: yes
   become_user: "{{ project_user }}"
 
-- name: check for secrets file
-  stat:
-    path: "{{ project_app_root }}/config/secrets.yml"
-  register: secret
```
rotated8, line 22: Was this variable being used? Does it need to be removed from anywhere else?

pmather: The original code was something left over from the version that Francis started.  I don't believe `secret` was ever actually used.  It is not needed now, at least.

rotated8: :+1:

PR #85, opened 2016-08-30
add a missible variable

kayiwa:
this fixes a "failed": true msg which asks for a remote user when
running this without vagrant

### DIFF
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -1,6 +1,8 @@
 ---
 # Commonly Changed Variables:
 #
+# Ansible User
+ansible_user: 'superuser'
 # Name of project application
 project_name: 'data-repo'
 # The project branch of the application to be used


pmather:
This variable used to be in `example_site_secrets.yml` but was removed because it is ignored (or rather overridden) by Vagrant and also is a bit misleading in the context of multiple providers (AWS, OpenStack, and VirtualBox) whose base boxes all use different values (i.e., `ubuntu`, `cc`, and `vagrant` respectively).

Given that, I believe it's confusing to list this in the `Commonly Changed Variables` section of `example_site_secrets.yml` because only people invoking Ansible directly (i.e., not via Vagrant) will ever need to set it.  It should be put in its own commented section elsewhere in the file (e.g., `# Variables related to direct Ansible invocation:`) and preferably commented out, to avoid possible bad interaction with Vagrant.

PR #90, opened 2016-09-01
call iawa:update_images rake task to add carousel images

whunter:
screen shot 2016-09-01 at 3 27 35 pm


ansible/roles/sufia/tasks/iawa.yml
```diff
@@ -0,0 +1,6 @@
+---
+- block:
```
pmather, line 2: FWIW, "block" is superfluous here for a single task, but not actually an error.

PR #91, opened 2016-09-01
Add Ansible role to install FITS

pmather:
This adds a "fits" role to install FITS for Sufia applications and adds
it as a dependency of the "sufia" role.  FITS is installed to /opt and
the version installed is symlinked to /opt/fits (as in the previous
BASH script InstallScripts).

Version 0.8.5 of FITS is installed for IAWA (i.e., Sufia 7); version
0.6.2 otherwise.  This can be overridden by setting "fits_version" to
the desired version in site_secrets.yml.

### DIFF
--- a/ansible/roles/fits/README.md
+++ b/ansible/roles/fits/README.md
@@ -1,0 +1,17 @@
+Fits
+====
+
+Installs FITS under /opt and symlinks the version installed to /opt/fits
+
+Requirements
+------------
+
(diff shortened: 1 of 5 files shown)


rotated8:
FITS 0.8.6 and 0.10.1 have slightly odd download filenames: fits-0.8.6_1.zip and fits-0.10.1_0.zip respectively. We should make a note of this.

PR #93, opened 2016-09-08
apt-key fails to verify SSL certificates

kayiwa:
when using the apt-key module with python less that 2.7.9 SSL
verification fails. Ubuntu 14.04 has python 2.7.6 which makes
installation fail.


kayiwa:
@rotated8 pull request is cleaner.

PR #96, opened 2016-09-14
Precompiles assets if the secrets change

rotated8:


rotated8:
Actually, this feels like a bad band-aid for the situation.

PR #100, opened 2016-09-30
Only installs Fedora, Solr, and Redis if they have local connections.

rotated8:
Connections need to be to 'localhost' or '127.0.0.1'. Support for IPv6 ('::1') is not implemented. Resque is still installed locally if it is present in the gems.


rotated8:
Much more refactoring is needed to conditionally install PostgreSQL.

rotated8:
Additionally, if Solr is installed remotely, the application cannot create a new core, so the core must already exist.

kayiwa (approved):
thanks!

PR #104, opened 2016-10-11
Allow Fedora Java VM options to be specified

pmather:
We introduce into the fedora4 role a new variable,
"fedora_java_vm_opts", that is used for specifying the VM-related
options (max heap size, etc.) to JAVA_OPTS.

This option is represented at the project level by the Ansible variable
"project_fedora_java_vm_opts".  By use of this variable, pre-production
and production servers can tailor the amount of memory available to
Fedora 4 via Tomcat.


rotated8:
Does this handle merging options if not all are defined?

rotated8 (approved):

PR #105, opened 2016-10-18
Add additional settings to define which ORCiD service is used

pmather:
Here we add extra fields to the "orcid" key of the config/secrets.yml
template so that either the "sandbox" or "production" ORCiD services
may be used.

Which settings are used is governed by the "project_orcid_service"
entry in site_secrets.yml.  URLs corresponding to the production
service are used if "project_orcid_service" is set to "production".
Otherwise, settings appropriate for the sandbox service are used.

### DIFF
--- a/ansible/roles/sufia/defaults/main.yml
+++ b/ansible/roles/sufia/defaults/main.yml
@@ -11,3 +11,8 @@ project_deploy_key: ''
 project_app_root: '{{ project_user_home }}/{{ project_name }}'
 # Name of Solr test core used by application
 project_solr_test_core: 'test'
+# ORCID API endpoints
+project_orcid_site_url: "{{ 'http://api.orcid.org' if project_orcid_service == 'production' else 'http://api.sandbox.orcid.org' }}"
+project_orcid_token_url: "{{ 'https://api.orcid.org/oauth/token' if project_orcid_service == 'production' else 'https://api.sandbox.orcid.org/oauth/token' }}"
+project_orcid_remote_signin_url: "{{ 'https://orcid.org/signin/auth.json' if project_orcid_service == 'production' else 'https://sandbox.orcid.org/signin/auth.json' }}"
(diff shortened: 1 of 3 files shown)


ansible/roles/sufia/defaults/main.yml
```diff
@@ -11,3 +11,8 @@ project_deploy_key: ''
 project_app_root: '{{ project_user_home }}/{{ project_name }}'
 # Name of Solr test core used by application
 project_solr_test_core: 'test'
+# ORCID API endpoints
+project_orcid_site_url: "{{ 'http://api.orcid.org' if project_orcid_service == 'production' else 'http://api.sandbox.orcid.org' }}"
```
rotated8, line 15: Does `"http://api.{{ 'sandbox.' if project_orcid_service == 'sandbox' }}orcid.org"` work?

vagrantlog.txt
```diff
@@ -0,0 +1,439 @@
+Script started on Tue Oct 18 14:24:53 2016
```
rotated8: Please delete this file

rotated8 (commented):
Got a couple of comments for you.

ansible/roles/sufia/defaults/main.yml
pmather: I didn't try that.  I'm loathe to try and follow a pattern like that because if the sandbox and production URLs differ radically in the future then any such pattern will break down.  I thought it best and most straightforward just to include the full URL for each service in the assignment.  The way it's done in the pull request covers the general case.

vagrantlog.txt
pmather: Sorry, will do.  That file sneaked in by accident.  Thanks for catching it.

ansible/roles/sufia/defaults/main.yml
pmather: I also wanted to be explicit in selecting the URLs for the production service so that the failure mode (if `project_orcid_service` wasn't set properly) would be to default to the sandbox service URLs.

rotated8 (approved):

PR #107, opened 2016-10-21
Gather local files under local_files

pmather:
We aim for the convention that locally-provided files should go under
the local_files directory in the Ansible playbook directory (ansible/).


rotated8 (changes requested):
Please add PhantomJS to local_files, and add notes to the sufia, tls_cert and phantomjs roles' READMEs. The files directories (or the .gitkeep files in them) can be deleted too.

rotated8 (approved):

PR #110, opened 2016-10-27
Adds caching to the FITS role

rotated8:


ansible/roles/fits/tasks/main.yml
```diff
@@ -7,6 +7,13 @@
   with_items:
     - unzip
 
+- name: copy cached fits
+  copy:
+    src: '{{ local_files_dir }}/fits-{{ fits_version }}.zip'
+    dest: /tmp/fits-{{ fits_version }}
```
pmather: Is this missing a ".zip" on the end?

pmather (commented):
I think this line is missing a ".zip" on the end of the destination.  Elsewhere in the tasks it is referred to as `/tmp/fits-{{ fits_version }}.zip`.

ansible/roles/fits/tasks/main.yml
rotated8: Good catch!

rotated8:
@pmather Updated.

PR #111, opened 2016-10-29
Prefer ansible over ansible_local

rotated8:
If Ansible is installed, use it, rather than the ansible_local provisioner.


pmather (approved):
I like that this change still lets you explicitly select ansible vs. ansible_local via the ANSIBLE_PROVISIONER environment variable.  I approve merging.

PR #112, opened 2016-11-04
Caches the NOID minter statefile [WIP]

rotated8:
The statefile is responsible for preventing ID collisions when creating objects. By caching it, we can avoid ID collisions. This is only really an issue for production and preprod.

### DIFF
--- a/ansible/roles/sufia/tasks/data-repo.yml
+++ b/ansible/roles/sufia/tasks/data-repo.yml
@@ -63,6 +63,18 @@
       dest: '{{ project_app_root }}/app/assets/images/carousel/'
     ignore_errors: True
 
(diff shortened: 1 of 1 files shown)


pmather (commented):
Would you explain a bit further the intention behind this change?  It appears to use "caching" in a difference sense to other InstallScripts pull requests (e.g. #79 and #110).

This change appears only to copy the installed `/tmp/minter-state` file from the installed system to the local deployment machine.  How will this prevent ID collisions?

rotated8:
Sorry, this is a WIP. Effectively, we should be backing this up so we can restore if needed.

rotated8:
This will be handled with other issues.

PR #113, opened 2016-11-09
Adds Noid minter statefile configuration

rotated8:


ansible/roles/sufia/tasks/main.yml
```diff
@@ -92,6 +92,28 @@
   when: resque_installed.stat.exists
 
 - block:
+  - name: ensure the Noid minter statefile dir exists
+    file:
+      path: '{{ project_noid_statefile.rsplit("/")[0] }}'
```
pmather: I suggest you use the `| dirname` filter instead

ansible/roles/sufia/tasks/main.yml
```diff
@@ -92,6 +92,28 @@
...
+    stat:
+      path: /tmp/minter-state
+    register: minter_state
+
+  - name: move old Noid statefile
+    copy:
+      src: /tmp/minter-state
+      dest: '{{ project_noid_statefile }}'
+      remote_src: True
+    when: minter_state.stat.exists
```
pmather: I don't think we should automagically move a file into the state file location if something exists in `/tmp/minter-state`.  (Given the permissive permissions on `/tmp`, this would provide an easy vector to sabotage the minter state file.)  I believe it is better to migrate the legacy file by hand as a one-time operation and not do it via a recurring task in Ansible.

pmather (changes requested):
Please consider these changes.  I believe using `dirname` is more standard, and that clobbering the minter state is not safe.

Also, please consider adding ownership and permissions to the task concerning the minter state directory.  This directory should be owned by the Rails application (`project_user`) and writeable by that user.

ansible/roles/sufia/tasks/main.yml
rotated8: Yes, I missed that filter.

ansible/roles/sufia/tasks/main.yml
rotated8: Ok.

rotated8:
The directory is created in a block that is run as the application user, so it should be created with the correct permissions already.

pmather (approved):

ansible/roles/sufia/tasks/main.yml
```diff
@@ -92,6 +92,11 @@
   when: resque_installed.stat.exists
 
 - block:
+  - name: ensure the Noid minter statefile dir exists
+    file:
+      path: '{{ project_noid_statefile | dirname }}'
+      state: directory
+
```
pmather, line 103: What do you think about adding a task to make sure the permissions and ownership of the minter state file are correct?  E.g.,

```
- name: ensure the Noid minter statefile has correct ownership and permissions
  file:
    path: '{{ project_noid_statefile }}'
    owner: '{{ project_user }}'
    mode: '0644'
    state: file
```

This wouldn't have any effect if the minter state file doesn't already exist, but it would serve to enforce correct permissions when it does, if those somehow get set incorrectly externally.  It also serves to make explicit what we expect the correct permissions to be for the minter state file.

rotated8: Ok, that works for me.

rotated8:
I realized that checking the ownership of the statefile should happen after the directory is ensured to exist, but in the case that the file is not owned by the application user, the application user would not have permission to change the ownership. So, I have opted to move the directory creation out of the application user block, and explicitly specify the user and group for the directory.

pmather:
I just realised that with the default for `project_noid_statefile` in `ansible/example_site_secrets.yml` this would have the intention of making the application `config/` directory explicitly writable (for the purposes of allowing the creation of the minter statefile), which is probably not a good idea (from a security point of view).  We should encourage setting `project_noid_statefile` such that it is outside the code base (along with fixing up directory permissions in our various application repositories).

This also brings an implicit responsibility when choosing a directory for `project_noid_statefile` as its ownership will be changed.

Maybe we could put runtime data in `/var/opt/{{ project_name }}` as a default?

rotated8:
With Sufia 7, moving the statefile should no longer be necessary. By default, the statefile will be at `/var/sufia/minter-state` when the application is in production mode, and `/tmp/minter-state` otherwise. We could adopt that as the default.

pmather (changes requested):
If the intention of this change is to make sure the minter state file directory is writeable by the application, you should add `mode: u=rwx` or something similarly explicit to the file task to ensure this.

rotated8:
Updated

pmather (changes requested):
I agree it's a good idea to follow the upcoming default for Sufia 7 regarding the location of the minter state file.

Given your comment about defaults for production vs. development, I suggest setting `project_noid_statefile` conditionally in the defaults for the role, e.g.,

```
project_noid_statefile: "{{ '/var/sufia/minter-state' if project_app_env == 'production' else '/tmp/minter-state' }}"
```

rotated8:
Given that the initializer in data-repo assumes `/tmp/minter-state` if the variable is empty, would `project_noid_statefile: "{{ '/var/sufia/minter-state' if project_app_env == 'production' }}"` be enough?

pmather:
There's an argument to be made to simply default it to `/var/sufia/minter-state` and put a warning saying "be sure you know what you're doing if you change this to something else."

Having it default to `/tmp/minter-state` (or encouraging the use of that location) in development mode also brings considerable risk given we mess with the ownership of the directory in which the minter state file resides.  For one, it would change ownership of `/tmp`, which may have very bad side effects.  That's why I believe if we explicitly set the ownership and permissions of the minter state file directory such that the Hydra application is guaranteed to be able to write to it, it is incumbent upon us to use a directory location for which that is absolutely safe to do.  I believe that is safe for a place like `/var/sufia`; I'm not so sure for places like `/tmp`, hence the warning/explanation about the ramifications of setting this.

So, I'm okay with just leaving your original change (defaulting to `/var/sufia/minter-state`), given the above caveats.

rotated8:
Squashed.

pmather (approved):

PR #114, opened 2016-11-09
Handlers run as root.

rotated8:


pmather (approved):
I've verified this as working for me.

PR #119, opened 2016-11-14
Includes project tasks only if they exist.

rotated8:


pmather (approved):
This change is tested to work correctly for me with both the "ansible" and "ansible_local" Vagrant provisioners.

PR #122, opened 2016-11-21
Updates README about Ansible provisioners

rotated8:


rotated8:
Thanks to @zoto724 for pointing this out.

pmather (approved):
Technically, "installed"/"not installed" means installed/not installed on the search path.  Ansible may be installed on the system, but if it is not on the search path then it won't be picked up by the `which` command.  Alternatively, you could use "detected to be installed" instead of the word "installed."

rotated8:
@pmather Updated.

pmather (approved):
That explanation is overly technical if not brutally accurate. 😄

PR #123, opened 2016-12-06
Stop installing Gems as root

pmather:


ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
```diff
@@ -24,24 +24,29 @@
...
     state: present
-    chdir: "{{ project_app_root }}"
-    exclude_groups: 'development test'
-    user_install: no
-  environment:
-    RAILS_ENV: '{{ project_app_env }}'
-  when: project_app_env == "production"
+    chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    gem_path: vendor/bundle
```
rotated8, line 81: I think we need `deployment_mode: False` here in case someone was using the production env and switches to development.

ansible/roles/sufia/tasks/main.yml
```diff
@@ -75,15 +75,27 @@
 
 - name: install the project's gems for development
   bundler:
+    state: present
     chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    gem_path: vendor/bundle
```
rotated8, line 128: Same here as above.

rotated8 (changes requested):
I seem to remember Bundler's deployment mode being sticky. I think we need to explicitly turn it off for development.

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
pmather: It appears that Ansible's `deployment_mode` behaves like [Bundler's `--deployment` flag](http://bundler.io/v1.13/deploying.html) and not like the description given in the [Ansible bundle module documentation](http://docs.ansible.com/ansible/bundler_module.html).  It's not clear to me what behaviour you're seeking with this ticket.  Note that `deployment_mode` is supposed to default to `no`.  I will add it explicitly, though.

ansible/example_site_secrets.yml
```diff
@@ -115,7 +119,7 @@ project_app_root: '{{project_user_home }}/{{ project_name }}'
 # Username for chroot SFTP user
 sftp_user: "upload"
 # Username that should be the primary owner of the chroot SFTP directory area
-sftp_chroot_owner: "{{ project_user }}"
+sftp_chroot_owner: "{{ project_run_user }}"
```
rotated8: I think I'm going to start treating changes to example_site_secrets.yml as breaking changes by default, and make changes to the file require a major version release as well. I don't want that to prevent changes to the file, but I want a better version number attached to this repo.

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
```diff
@@ -6,7 +6,6 @@
     owner: "{{ project_user }}"
     group: "{{ project_group }}"
     state: directory
-    recurse: yes
```
rotated8, line 9: We don't want to recurse because that means the files that must be owned by the project_run_user will be changed to the project_user and back right? But if we aren't recursing, this only handles the directory. If we don't have problems with the rest of the application, should this task be removed in another ticket?

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
```diff
@@ -15,42 +14,116 @@
...
-    chdir: "{{ project_app_root }}"
-    exclude_groups: 'development test'
-    user_install: no
-  environment:
-    RAILS_ENV: '{{ project_app_env }}'
-  when: project_app_env == "production"
+    chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    deployment_mode: no
+    gem_path: vendor/bundle
```
rotated8, line 81: Do we need to ensure this directory exists?

ansible/roles/sufia/tasks/main.yml
```diff
@@ -14,6 +14,14 @@
     - libsqlite3-dev
     - zlib1g-dev
 
+- name: make sure project user owns application
```
rotated8: Same question as above about necessity of this task.

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
```diff
@@ -15,42 +14,116 @@
...
     state: present
-    chdir: "{{ project_app_root }}"
-    user_install: no
-  environment:
-    RAILS_ENV: '{{ project_app_env }}'
-  when: project_app_env != "production"
+    chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    deployment_mode: no
+    gem_path: vendor/bundle
```
rotated8, line 95: Do we need to ensure this directory exists?

ansible/roles/sufia/tasks/main.yml
```diff
@@ -56,38 +66,90 @@
...
-    state: directory
-    recurse: yes
-
 - name: install the project's gems for development
   bundler:
+    state: present
     chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    deployment_mode: no
+    gem_path: vendor/bundle
```
rotated8, line 128: Do we need to ensure this directory exists?

ansible/roles/sufia/tasks/main.yml
```diff
@@ -56,38 +66,90 @@
...
+  become: yes
+  become_user: '{{ project_user }}'
 
 - name: install the project's gems for production
   bundler:
+    state: present
     chdir: '{{ project_app_root }}'
+    binstub_directory: bin
+    deployment_mode: no
+    gem_path: vendor/bundle
```
rotated8, line 142: Do we need to ensure this directory exists?

rotated8 (changes requested):
Couple of questions (two, really) for you as I work on renaming the users.

ansible/example_site_secrets.yml
pmather: You'll have to define better what is considered a breaking change.  Here, I'm just changing the value of an `example_site_secrets.yml` variable, not changing the name or deleting an existing variable.  However, the change was introduced because of the change in semantics due to the new run/owner users that were introduced in this change.  Those are definitely breaking semantics compared to the previous version of `master`.

So, having said that, I'm happy for us to bump the semantic version and tag a new release every time we have non-comment changes to `example_site_secrets.yml`.

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
pmather: Bundler creates this directory.

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
pmather: Bundler creates this directory.

ansible/roles/sufia/tasks/main.yml
pmather: Bundler creates this directory.

ansible/roles/sufia/tasks/main.yml
pmather: (See above.)

ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
pmather: I removed the `recurse: yes` because this applied the permissions recursively, too.  You are correct in your comment about `recurse: yes` also changing ownerships when we don't really want this. This change also goes hand in hand with the `become_user: "{{ project_user }}"` on the Git clone operation, which is also new.  That causes the directories to be owned by `{{ project_user }}`, something the `recurse: yes` was originally included to achieve.

This way, we can be more flexible about setting the directory permissions in the code repo (which Git does maintain, IIRC), which may be a bit easier that adding a lot of `file:` blocks to be very explicit about their settings.

In this set of changes, I tend to reserve the `file:` blocks for those cases when the ownership and permissions should be assigned to the `{{project_run_user}}`.

ansible/roles/sufia/tasks/main.yml
pmather: I did this to be parallel to the way it was done in the `geoblacklight` role.

rotated8:
Updated: 0af59a56379315e5195f4970b7e2d93d9d35eaf9

rotated8 (approved):

zoto724 (commented):
Is the desired order of operations:
1.  approve, then test and make changes on dev if necessary?
or
2. test branch on dev, then approve if it works?

Either is fine with me, this looks good.

pmather:
I would say 2), i.e., only approve if you believe the proposed pull request changes work and that the pull request can be safely merged into master.  Testing on dev (or Vagrant) first and then approving if it works would be a good way about doing that.

rotated8:
Yes, my preference is for 2 as well.

ansible/roles/sufia/tasks/data-repo.yml
```diff
@@ -64,6 +64,6 @@
     ignore_errors: True
 
   become: yes
-  become_user: '{{ project_user }}'
+  become_user: '{{ project_runner }}'
```
rotated8: @pmather, can you talk about why this is `project_runner` and not `project_owner`? The images added above are (currently) owned by hydra, so they can't be deleted by railsapps. @zoto724 ran into this yesterday.

pmather: It probably made sense to me to use `project_runner` there as that is the user under which the Rails app will ultimately run.  However, I now believe this is likely incorrect and a victim of a lack of testing on my part.  Lots of the actions done in the `block` have `ignore_errors: True` and are optional.  I don't have a `user_list.txt` nor an `admin_list.txt` and so didn't test those code paths.  Also, I don't have any carousel images `images.zip` file and so wasn't able to test that code path either.

Looking at the `block` of actions, it does seem more reasonable to execute these as `become_user: '{{ project_owner }}'`.  Did that work for you the other day?

rotated8: Yes. I have a commit for this in data-repo.yml and iawa.yml. I don't know if this needed for geoblacklight. I'll squash all the commits while I'm at it.

pmather: It looks like geoblacklight is using `project_owner` as appropriate.

I will need to get a sample `user_list.txt`, `admin_list.txt` and `images.zip` from you for future testing.

rotated8:
Updated again.

zoto724:
Looks good with this change

PR #124, opened 2016-12-06
Fix missing handler

pmather:
With the release of Ansible 2.2, handlers were changed explicitly to
run as root when restarting system services.  The handler in the
chroot-sftp role was inadvertently left out in those updated as part of
pull request #114.  This commit rectifies that.


rotated8:
Probably my fault. I changed the ones I knew from Sufia, and didn't check GBL.

rotated8 (approved):

PR #126, opened 2016-12-19
Support local or remote PostgreSQL servers for Rails DB

pmather:
The original BASH scripts supported the use of a local or remote
PostgreSQL server for the Rails ActiveRecord DB.  The Ansible version
was a regression in that it supported only the use of a local
PostgreSQL server.  This commit eliminates this regression and allows
either a local or a remote PostgreSQL server to be used.

If {{project_db_host}} points either to "localhost" or a directory (in
which the PostgreSQL Unix socket resides) then it is assumed a local
PostgreSQL server is to be used.  In this case a local server is
installed and set up.  Otherwise, it is assumed a remote PostgreSQL
server (already installed) is to be used, in which case only necessary
PostgreSQL client libraries are installed.

### DIFF
--- a/ansible/roles/postgresql/defaults/main.yml
+++ b/ansible/roles/postgresql/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+# PostgreSQL-related settings
+postgresql_is_local: "{{ 'true' if project_db_host | search( 'localhost|127\\.0\\.0\\.1|^/') else 'false' }}"
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -61,6 +61,12 @@ project_db_name: 'datarepo'
 project_db_user: '{{ project_user }}'
 # Application database name for Rails config/database.yml "password:" setting
 project_db_password: 'changeme'
+# Host of PostgreSQL server
+project_db_host: '/var/run/postgresql'
+# PostgreSQL admin user name for non-local db creation
+project_db_admin_user: postgres
+# PostgreSQL admin password for non-local db creation
+project_db_admin_password: postgres
 # Base URL of Solr endpoint
 project_solr_url: 'http://localhost:8983/solr'
 # Name of Solr core used by application
--- a/ansible/roles/postgresql/README.md
+++ b/ansible/roles/postgresql/README.md
@@ -11,8 +11,12 @@ Apt must be installed, along with our Common role.
 Role Variables
 --------------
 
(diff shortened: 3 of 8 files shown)


ansible/roles/postgresql/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# PostgreSQL-related settings
+postgresql_is_local: "{{ 'true' if project_db_host | search( 'localhost|127\\.0\\.0\\.1|^/') else 'false' }}"
```
rotated8, line 3: Took me a minute to see ^/ here. We should probably add the IPv6 loopback ::1 too.

rotated8: Also can you talk about why you're using strings instead of Python's booleans `True` and `False`?

rotated8 (changes requested):

ansible/roles/postgresql/defaults/main.yml
pmather: The `search` test was lifted from the `when:` clauses in `ansible/sufia_site.yml`, i.e., it uses the same "local" criteria as used to decide whether to install fedora, solr, or redis.  If we add the IPv6 loopback here, it probably should be added to those, too.

I used 'true' because that seems to be used a lot elsewhere in our roles.  However, after reading your comment, I notice the Python 'True' is also used in a few places, and our usage is inconsistent.

I don't know whether or not Python booleans are more appropriate in Ansible playbooks.  Although Ansible is implemented in Python and uses Jinja2, it is a DSL, not actually Python code.  The Ansible module documentation tends to use `yes` or `no` when describing parameters that take on boolean values, so maybe that is more appropriate?  The [Ansible YAML syntax](https://docs.ansible.com/ansible/YAMLSyntax.html) says that many alternative forms are acceptable for boolean True/False.

The [Jinja2 template documentation on literals](http://jinja.pocoo.org/docs/dev/templates/#literals) actually explicitly prefers `true` and `false` instead of `True` and `False` and indeed states that in previous versions of Jinja2 that `True` and `False` evaluated to and undefined variable (i.e., `false`) but that is no longer the case because of the confusion it caused.  Because the statement line is Jinja2 (because of the {{...}}), arguably `true` and `false` are correct.

rotated8 (approved):
Ok. I see two tickets coming out of this: one for standardizing boolean handling, and one for adding IPv6 support.

PR #127, opened 2017-02-03
Add task to establish initial controlled vocabs in Iawa

tingtingjh:


pmather (commented):
Is the `iawa:add_controlled_vocabs` rake task idempotent?  I.e., can you run it multiple times in succession and still have it just add one set of controlled vocabulary terms?  Alternatively, will running it a second time cause an error because the controlled vocabulary created in the first run is already present?

tingtingjh:
Thanks for pointing this out. Now it should be an idempotent task.

PR #128, opened 2017-02-08
make public/uploads directory writable so file upload will work

whunter:


pmather (commented):
It was my impression talking to Collin that the preferred/better solution was to move uploads to the application tmp directory.  The proposed change in this pull request conflicts with that strategy.

ansible/roles/sufia/tasks/main.yml
```diff
@@ -89,6 +89,8 @@
       mode: "0770"
     - path: "{{ project_app_root }}/db"
       mode: "0770"
+    - path: "{{ project_app_root }}/public/uploads"
```
pmather, line 92: Note that `config/initializers/sufia.rb` for both `data-repo` and `iawa` currently contain this:
```
config.upload_path = ->() { Rails.root + 'tmp' + 'uploads' }
```

So, that resultant path is the one that needs to be writeable.  From where in the application code are you getting the path `{{ project_app_root }}/public/uploads`?

pmather (commented):
Note that `config/initializers/sufia.rb` for both `data-repo` and `iawa` currently contain this: `config.upload_path = ->() { Rails.root + 'tmp' + 'uploads' }`.  So, that resultant pa

ansible/roles/sufia/tasks/main.yml
whunter: Even with the tmp/uploads path being assigned to the upload_path variable the application is still trying to write files to public/uploads. We've yet to figure out why.

pmather: It would be nice to try and figure out what is not working with the `config.upload_path` setting or what we're possibly misunderstanding about it.  (The `make_pub_uploads_writable` branch is a handy workaround until that happens, but I'm not comfortable with merging this to `master` until we get a better understanding.)

The `config.upload_path` setting is used here: https://github.com/projecthydra/sufia/blob/master/app/uploaders/sufia/uploaded_file_uploader.rb

You'll note there that was added as part of a commit that fixes projecthydra/sufia#2083. The commit message begins, "Put uploaded files in a location where they will not be publically available." The aforementioned Sufia issue explains why they believe it is a problem to upload to the `public/` directory, and includes a link to the `CarrierWave` wiki on doing uploads securely.

If `config.upload_path` (added as a milestone for 7.0.0 with the projecthydra/sufia#2083 issue) is not being honoured, we should at least report this as a bug to Sufia.  If we need to add this to support pre-7.x Sufia then we should at least add a deprecation comment to `InstallScripts` with a milestone of removing it when we move to Sufia 7.x for everything.

(There may be a good case to make explicit the version of Sufia we are targeting in the Ansible `sufia` role so we can add logic to handle these corner cases properly in the future.)

PR #129, opened 2017-02-18
Provides more descriptive vm name for Virtualbox

whunter:
Allows user to specify vm name for Virtualbox and falls back to "project_name" if not defined. Useful for me so that I can identify my vm's in the Virtualbox gui.
Currently, all vm's are named similarly.
before
This change allows them to be more descriptive
after

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -84,6 +84,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     hydravm.vm.hostname = "hydravm"
 
     hydravm.vm.provider :virtualbox do |vb, override|
+      vb.name = $secrets_items["vm_name"] if !$secrets_items["vm_name"].nil? && !$secrets_items["vm_name"].empty?
+      vb.customize ["modifyvm", :id, "--description", "Created from Vagrantfile in #{Dir.pwd}"]
       override.vm.box = "ubuntu/trusty64"
       vb.memory = 4096
       vb.cpus = 2
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -1,6 +1,8 @@
 ---
 # Commonly Changed Variables:
 #
+# VM Name
+# vm_name: 'testing'
 # Name of project application
 project_name: 'data-repo'
 # The project branch of the application to be used


pmather (changes requested):
As someone who almost never looks at the VirtualBox GUI nor uses it to manage Vagrant machines, I am struggling to appreciate the need for this change. 😃 

Having said that, if you are to press ahead with this change, please make it entirely optional.  In other words, only set `vb.name` in the `virtualbox` provider if it is _explicitly_ requested in `site_secrets.yml` by having `vm_name` set to something.  Otherwise, do not set `vb.name` at all, which preserves the current behaviour.

One of the reasons setting the machine name was removed in the first place was that (IIRC) it does not allow multiple versions of the same VM to defined at the same time on the same machine.  Thus, with the above change, you would not be able to have two `iawa` VMs at the same time (even if they weren't running).  After getting an error to this effect when trying to create the second `iawa` VM, you'd have to explicitly rename it in `site_secrets.yml` before you could do so.  That, IMHO, is unnecessary extra work for those who always interact with Vagrant boxes using the `vagrant` command (which is ostensibly the "Vagrant way"), and never see what is listed inside VirtualBox.

I argue that it should be those who actually care what the VirtualBox GUI calls their VMs that should have the onus of specifying that name, not the other way around.

whunter:
@pmather This has been updated to reflect your change request. I generally use the command line to interact with vagrant also but sometimes I'll have quite a few vm's created that I would like to clean up and I lose track of which one belongs to which directory.

pmather (changes requested):
I still think this only partially helps with locating which VirtualBox VM is where.  (I imagine you'd get to the state where you have `iawa_1`, `iawa_2`, etc. VMs listed in the VirtualBox GUI and still have little idea of where they are actually located on disk.)

For our setup, looking at the Settings -> Shared Folders path for `vagrant` will let you know where the VM project lives.  Thus, a preferable change may be to use the `customize` method (described in the "VBOXMANAGE CUSTOMIZATIONS" section of the [VirtualBox provider customisation page](https://www.vagrantup.com/docs/virtualbox/configuration.html)) to add content to the `description` metatdata for the VM.  E.g.,
```
v.customize ["modifyvm", :id, "--description", "Created from Vagrantfile in #{vagrantfile_dir}"]
```
or something similar (where `vagrantfile_dir` would be code representing the dirname of the Vagrantfile used in `vagrant up`).

This would result in Settings -> General -> Description containing useful information locating the VM.

This idea could also be extended to include other useful metadata about the VM created, e.g., the time `vagrant up` was executed, etc., so that such metadata was contained all in one place.  This could be a general place for project-related metadata about VMs.

I have no objections to your `vb.name` change, but please also add the `--description` customisation feature.  This could either be as an addition to `vb.name` or as a replacement for it.

whunter:
@pmather I've added the description customization. Seems to work as expected.

pmather (approved):
Thanks for this change.

PR #130, opened 2017-02-20
add rake task to load workflows to data-repo tasks

whunter:
Loads default workflows from json files into db

### DIFF
--- a/ansible/roles/sufia/tasks/data-repo.yml
+++ b/ansible/roles/sufia/tasks/data-repo.yml
@@ -63,6 +63,12 @@
       dest: '{{ project_app_root }}/app/assets/images/carousel/'
     ignore_errors: True
 
+  - name: add workflows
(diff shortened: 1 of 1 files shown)


ansible/roles/sufia/tasks/data-repo.yml
```diff
@@ -63,6 +63,12 @@
       dest: '{{ project_app_root }}/app/assets/images/carousel/'
     ignore_errors: True
 
+  - name: add workflows
+    command: ./bin/rails curation_concerns:workflow:load
+    args:
+      chdir: '{{ project_app_root }}'
+    ignore_errors: True
```
pmather, line 70: Why are errors being ignored here?

pmather (commented):
It is often unusual to ignore errors unilaterally (e.g., not to `register` the result and then have subsequent tasks which check/handle the failure you expect to happen).  Is an unexamined `ignore_errors` okay here?

PR #133, opened 2017-09-22
add zip package from apt

whunter:
I'm working on a feature for VTechData which would let users download all items in a dataset in a zip file. I tried using the rubyzip gem but it modifies the files in some way that affects the checksums, which is not desirable.

### DIFF
--- a/ansible/roles/sufia/tasks/main.yml
+++ b/ansible/roles/sufia/tasks/main.yml
@@ -7,6 +7,7 @@
     update_cache: yes
   with_items:
     - git
+    - zip
     - unzip
     - imagemagick
     - libreoffice


pmather:
Just out of interest, have you considered using the [archive](https://rubygems.org/gems/archive/versions/0.0.6) Gem instead?  (See [its home page](https://github.com/erikh/archive) for details of how to install and use it.)  It uses `libarchive`, which is cross-platform and supports many compression and archive formats besides just ZIP.  You would likely need to add the `libarchive-dev` package instead of `zip` in order to build the Gem.

whunter:
I'll take a look at that one in the morning. I wasn't aware of it. Rubyzip
was the only gem that seemed to be showing up in my searches.

PR #135, opened 2017-10-24
copy fonts archive from local_files on host to assets in app

whunter:


ansible/roles/sufia/tasks/data-repo.yml
```diff
@@ -63,6 +63,31 @@
...
+    file:
+      path: '{{ project_app_root }}/app/assets/fonts'
+      state: directory
+
+  - name: find old font files
+    find:
+      path: '{{ project_app_root }}/app/assets/fonts/'
+    register: fonts
+
+  - name: delete old font files
```
pmather, line 78: If we are deleting contents from an `assets` directory, should we not also notify a handler to run the `tmp:clear` rake task to clear the cache?  I seemed to have problems when flipping between having and not having the fonts in `fonts.zip` installed during provisioning.  This appeared to be an application caching issue rather than a browser caching issue for me.

ansible/roles/sufia/tasks/data-repo.yml
```diff
@@ -63,6 +63,31 @@
...
+    file:
+      path: '{{ item.path }}'
+      state: absent
+    with_items: '{{ fonts.files }}'
+
+  - name: add new font files
```
pmather: Similar to the above, if we are adding fonts then do we need to notify a handler to clear the asset cache via the `tmp:clear` rake task or otherwise notify the handler to restart the application?

pmather (commented):
We probably need to add a handler to `ansible/roles/sufia/handlers/main.yml` to clear the application cache due to adding/removing items from `app/assets/fonts/`.  This handler should be notified whenever assets are added/removed.

Other than that, the change is tested as working for me, both in the case where `fonts.zip` is present and when it is not.

PR #143, opened 2018-01-30
Enable support for Google Analytics ID

pmather:
This change restores the ability to supply a Google Analytics ID to a Sufia or Hyrax application.

The Google Analytics ID for the application is defined in "site_secrets.yml" via the "project_google_analytics_id" variable. The "project_google_analytics_id" variable should be commented out if the application is not to enable Google Analytics.


whunter (approved):
Works as expected. Thank you!

PR #144, opened 2018-02-01
assets need to be in place before precompile

whunter:
When running in production precompile seems to be running before the carousel images and fonts are copied into place resulting in multiple 404's. You can see the order in the vagrant/ansible output here: https://gist.github.com/whunter/35058f7bced3ca660028f6e7e0d108bd

We need the images and fonts to be in place before precompile is run like this:
https://gist.github.com/whunter/bc2664b1d138cc77e485ddf6d32810af

### DIFF
--- a/ansible/roles/sufia/handlers/main.yml
+++ b/ansible/roles/sufia/handlers/main.yml
@@ -1,19 +1,21 @@
 ---
-- name: precompile assets
-  command: bundle exec rake assets:precompile
+- name: clear rails cache
+  command: bundle exec rake tmp:cache:clear
   args:
     chdir: '{{ project_app_root }}'
-  when: project_app_env == 'production'
-  notify: restart nginx
+  notify:
+    - precompile assets
+    - restart nginx
   become: yes
   become_user: '{{ project_owner }}'
   environment:
     RAILS_ENV: '{{ project_app_env }}'
 
-- name: clear rails cache
-  command: bundle exec rake tmp:cache:clear
+- name: precompile assets
+  command: bundle exec rake assets:precompile
   args:
     chdir: '{{ project_app_root }}'
+  when: project_app_env == 'production'
   notify: restart nginx
   become: yes
   become_user: '{{ project_owner }}'


pmather (changes requested):
I believe there was a bug in the carousel/font handling before inasmuch as it explicitly triggered a clearing of the Rails cache for these assets but did not trigger a precompile of them for `RAILS_ENV=production`.  (I believe it's a bug because anything altering the `assets` directory during deployment should precipitate an `assets:precompile`, right?)

A quick fix for this would be to amend the `clear rails cache` handler so it also calls the `precompile assets` handler.

I would rather not fix this the way it is done in this pull request.  By bringing this code out of the `data-repo`-specific tasks it breaks the generic nature of the `sufia` role.  With this change, every `sufia` application is deemed to have an image carousel and specialised fonts, which is not the case.  Handling for those pertains only to `data-repo`.

The downside of the quick fix I propose is that it leads to the `precompile assets` handler being run twice.  That is an inefficiency, not an error, but it does lead me to an idea for perhaps a longer-term/alternative fix for this.

I believe the underlying problem is that the per-project tasks executed via the `include project specific tasks` Ansible task have been semantically overloaded too much.  Initially, these were tasks that needed to be performed at a system level to ensure the packages/directories/etc. were in place for the application.  That argued for it being run earlier in the task sequence.

Later, application-related Rake tasks were added that required a fully-running application to be deployed in order to complete correctly.  That argued for it being run later in the task sequence (where it now resides).

I believe that these two aspects (system-level vs. application-level) argue for two separate per-project task files to be included.  The system-related one (e.g., putting files in place) can be run earlier and the running application-related one can be run at the end.

For now, I can push my proposed quick fix to this pull request and we can discuss the longer-term fix at the next meeting?

whunter:
@pmather I think your system / application level split makes a lot of sense. In the interim though I'm fine with the precompile running twice. If you don't have any objections, I'll just go ahead and merge your changes. They seem to work fine.

PR #145, opened 2018-02-23
add fonts to compel/binaryArts application

whunter:


ansible/roles/hyrax/tasks/binary-arts.yml
```diff
@@ -26,6 +26,29 @@
...
+    find:
+      path: '{{ project_app_root }}/app/assets/fonts/'
+    register: fonts
+
+  - name: delete old font files
+    file:
+      path: '{{ item.path }}'
+      state: absent
+    with_items: '{{ fonts.files }}'
+    notify: clear rails cache
```
pmather, line 44: The `clear rails cache` handler is missing from `ansible/roles/hyrax/handlers/main.yml`, which causes this play to fail.  The `clear rails cache` entry from `ansible/roles/sufia/handlers/main.yml` could be prepended to `ansible/roles/hyrax/handlers/main.yml` to achieve the same effect as a quick way of fixing this.

pmather (changes requested):
You are referencing a handler (`clear rails cache`) that does not exist during this playbook run.

ansible/roles/hyrax/tasks/binary-arts.yml
whunter: @pmather I added that handler to hyrax.

pmather (approved):

PR #146, opened 2018-02-23
Obtain files for local_files from a remote URL

soumikgh:
Since we had issues with Ansible Tower not working well with extra files in a cloned repository, one of the solutions is to fetch these files from a remote URL and populate the local_files with these. This would otherwise have to be done by SSHing into the AWX node, going inside the awx_task Docker container and SCPing multiple times. (What I usually used to do is put the local_files.tar.gz on a web server somewhere and wget that there, instead of multiple SCPs, which is what this change will essentially do).

This will also get us closer to giving devs access to Ansible Tower, and them re-deploying the development instances by themselves. I'm going to set up access for the cPanel account which hosts hydrafiles.vtlibraries.net for the developers, so that they can make changes to the archives and deploy development instances. This would also help consolidate the extra files for the different applications.

Note: The role deletes and re-creates the local_files directory to empty it. There's a planned feature for the file module which will allow the directory to be emptied (https://github.com/ansible/ansible/issues/18910).

### DIFF
--- a/ansible/roles/hyrax/meta/main.yml
+++ b/ansible/roles/hyrax/meta/main.yml
@@ -11,3 +11,4 @@ dependencies:
   - { role: nodejs, nodejs_version: '6.x' }
   - { role: phantomjs, phantomjs_version: '2.1.1', when: project_app_env != 'production' }
   - { role: ruby, ruby_version: '2.3' }
+  - { role: local_files }
--- a/ansible/roles/local_files/tasks/main.yml
+++ b/ansible/roles/local_files/tasks/main.yml
@@ -1,4 +1,12 @@
 ---
+- name: delete the local_files directory (Only way is a delete and re-create until state=empty is implemented)
+  file: path="local_files/" state=absent
+  delegate_to: localhost
+  
+- name: recreate the local_files directory
+  file: path="local_files/" state=directory
+  delegate_to: localhost
+  
 - name: extract remote archive to local_files directory
   unarchive: copy=no src="{{ local_files_archive }}" dest="local_files/"
   when: local_files_archive is defined
--- a/ansible/roles/local_files/tasks/main.yml
+++ b/ansible/roles/local_files/tasks/main.yml
@@ -1,0 +1,15 @@
+---
+- name: delete the local_files directory (Only way is a delete and re-create until state=empty is implemented)
+  file: path="local_files/" state=absent
+  when: local_files_archive is defined
+  delegate_to: localhost
+  
+- name: recreate the local_files directory
+  file: path="local_files/" state=directory
+  when: local_files_archive is defined
+  delegate_to: localhost
+  
+- name: extract remote archive to local_files directory
+  unarchive: copy=no src="{{ local_files_archive }}" dest="local_files/"
+  when: local_files_archive is defined
+  delegate_to: localhost
--- a/ansible/roles/local_files/tasks/main.yml
+++ b/ansible/roles/local_files/tasks/main.yml
@@ -3,13 +3,16 @@
   file: path="{{local_files_dir}}/" state=absent
   when: local_files_archive is defined
   delegate_to: localhost
+  become: no
   
 - name: recreate the local_files directory
   file: path="{{local_files_dir}}/" state=directory
   when: local_files_archive is defined
   delegate_to: localhost
+  become: no
   
 - name: extract remote archive to local_files directory
   unarchive: copy=no src="{{ local_files_archive }}" dest="{{local_files_dir}}/"
   when: local_files_archive is defined
   delegate_to: localhost
+  become: no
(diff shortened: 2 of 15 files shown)


ansible/roles/local_files/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: extract remote archive to local_files directory
+  unarchive: copy=no src="{{ local_files_archive }}" dest="local_files/"
```
pmather: There is actually a site-wide variable, `local_files_dir` (defined in `ansible/site_vars.yml`) that designates the folder used for storing local files.  You should thus use `dest="{{ local_files_dir }}"` instead of hard-coding `dest="local_files/"`.

ansible/roles/hyrax/meta/main.yml
```diff
@@ -11,3 +11,4 @@ dependencies:
   - { role: nodejs, nodejs_version: '6.x' }
   - { role: phantomjs, phantomjs_version: '2.1.1', when: project_app_env != 'production' }
   - { role: ruby, ruby_version: '2.3' }
+  - { role: local_files }
```
pmather, line 14: The `local_files_dir` local files directory mechanism is utilised in more places than just the `sufia` and `hyrax` roles.  If you are wishing to populate local files from a remote archive then you will need to include the role in more places than those.

By my estimate, the local files directory is also used in these roles: `phantomjs`; `tls_cert`; `chroot-sftp`; and `fits`.

pmather (changes requested):
This change looks good except for the specific comments I have attached to the code elsewhere (i.e., using `local_files_dir` and increasing the role dependencies coverage).

As a general comment, in future it would be preferable to squash the initial pull request commit into a single unified commit.  This helps eliminate unnecessary commits (such as the commit that just reverts a previous commit) and helps make clearer what actually has been changed (rather than illuminating the road to how you got there).  Once review begins on the pull request then it's appropriate to push additional commits, but in the DLD workflow we prefer to begin a pull request with a squashed single commit.

Regarding the role dependencies to enable the `local_files` role, make sure that however you do it, the `local_files` role is enabled as early as possible.  (I think how you are doing that should ensure that.)

ansible/roles/hyrax/meta/main.yml
soumikgh: From what I understand, those roles use files ('cached' files) which are already available in the `local_files_dir`, and as far as I can see, those roles are always called from the hyrax_site and sufia_site roles (or other super-roles).

So, what I imagine is the proper course of action, is to put the `local_files` role before those roles as they are included in the hyrax_site and sufia_site roles. Or instead of explicitly including it in the super-roles, I can add it as a dependency for whichever roles which use `local_files_dir` (easily found out using `grep -r local_files_dir`).

Is my understanding of this correct?

ansible/roles/local_files/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: delete the local_files directory (Only way is a delete and re-create until state=empty is implemented)
+  file: path="{{local_files_dir}}/" state=absent
+  when: local_files_archive is defined
+  delegate_to: localhost
```
pmather, line 5: When using `vagrant` I need to add a `become: no` to make the `delegate_to: localhost` work.

ansible/roles/local_files/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  
+- name: recreate the local_files directory
+  file: path="{{local_files_dir}}/" state=directory
+  when: local_files_archive is defined
+  delegate_to: localhost
```
pmather, line 11: See above comment about using `become: no` with `vagrant`.

ansible/roles/local_files/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  
+- name: extract remote archive to local_files directory
+  unarchive: copy=no src="{{ local_files_archive }}" dest="{{local_files_dir}}/"
+  when: local_files_archive is defined
+  delegate_to: localhost
```
pmather, line 17: See previous comment.

pmather (changes requested):
This looks like it is almost there.  For it to work for me with `vagrant` I had to add a `become: no` to all the `delegate_to: localhost` actions.  (We do this in the `sufia` and `hyrax` `tasks/main.yml`, too.)

If adding `become: no` won't break things for your regular Ansible invocations (which I suspect it won't, due to us using it in the `sufia` and `hyrax` roles), then would you amend the `delegate_to` accordingly in `roles/local_files/tasks/main.yml`?  That will ensure the developers can use this with `vagrant`.

ansible/roles/local_files/tasks/main.yml
soumikgh: Works without issues on Tower. Added `become:no` to all the local actions in the new role.

pmather (approved):
This fixes the issue I was having using this change with Vagrant. 👍

PR #148, opened 2018-03-08
copy my personal .gitconfig file 

whunter:
to the same directory that the app sits in so that I don't always have to set this stuff manually. This is just a convenience thing for me, so if this isn't OK just let me know.

### DIFF
--- a/ansible/roles/hyrax/tasks/main.yml
+++ b/ansible/roles/hyrax/tasks/main.yml
@@ -230,6 +230,14 @@
   when: project_tasks.stat.exists
 
 - block:
+  - name: copy gitconfig
+    copy:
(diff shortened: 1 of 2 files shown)


pmather (commented):
I think that instead of embedding these in various roles, we should have a separate `creature_comforts` role that handles these miscellaneous "developer-friendly" actions.  The `creature_comforts` role could be run optionally at the end of a playbook run to copy various files or do other things as-needed, and need to be explicitly enabled via a `site_secrets.yml` setting (or else only run when `project_app_env` is `development`).

If you agree with the `creature_comforts` role idea, let me know and I'll work on it.

whunter:
Yeah, I think that makes perfect sense. Great idea.

PR #152, opened 2018-03-26
Pass in site_secrets.yml to Vagrant

pmather:
Currently, the site_secrets.yml vars file that controls much of the application provisioning is a mandatory file that is provided by the person doing the provisioning. Such externally-supplied files are not ideal when using Ansible Tower, which prefers a "pristine" environment in which to deploy, with optional components being supplied by Ansible Tower or fetched as part of the provisioning process.

To make Vagrant act similarly, we pass in the "site_secrets.yml" file explicitly as part of the provisioning process, similar to Ansible Tower's operation.  This is done by way of Vagrant's "extra_vars" mechanism.


soumikgh (approved):
Looks good to me! I was running off a `tower-deploy` branch, but with this merged, I can run off `master`. Thanks!

PR #153, opened 2018-04-10
add osf client details to secrets file

whunter:
used to inject the client id and the client secret into the application.

### DIFF
--- a/ansible/roles/sufia/defaults/main.yml
+++ b/ansible/roles/sufia/defaults/main.yml
@@ -20,5 +20,9 @@ graylog_protocol: udp
 graylog_network_locality: WAN
 graylog_facility: "{{ project_name }}"
 graylog_verbosity: "info"
+# OSF API settings
+osf_client_id: ""
+osf_client_secret: ""
 # Google Analytics
 sufia_google_analytics_comment_marker: "{{ '' if project_google_analytics_id is defined else '#' }}"
+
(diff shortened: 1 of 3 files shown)


ansible/roles/sufia/defaults/main.yml
```diff
@@ -20,5 +20,9 @@ graylog_protocol: udp
 graylog_network_locality: WAN
 graylog_facility: "{{ project_name }}"
 graylog_verbosity: "info"
+# OSF API settings
+osf_client_id: ""
```
pmather: Shouldn't this be `project_osf_client_id`?

ansible/roles/sufia/defaults/main.yml
```diff
@@ -20,5 +20,9 @@ graylog_protocol: udp
...
+osf_client_secret: ""
```
pmather: Shouldn't this be `project_osf_client_secret`?

pmather (changes requested):
The default variables for the `sufia` role are not used in the template or anywhere else.  They should be changed as suggested.

ansible/roles/sufia/defaults/main.yml
whunter: @pmather Change made

ansible/roles/sufia/defaults/main.yml
whunter: Change made

ansible/roles/sufia/templates/secrets.yml.j2
```diff
@@ -46,7 +46,11 @@ default: &default
   # secret_key_base is typically set to a long, random string, such as the output
   # of "openssl rand -hex 64" (or the output of the "rake secret" task).
   secret_key_base: abad1dea
-
+  # OSF client secrets
+  osf:
```
tingtingjh, line 50: osf: &osf?

tingtingjh (approved):

PR #156, opened 2018-05-03
Use Xenial base image

soumikgh:
Ubuntu 14.04, which is going EOL this year, is currently being used as the base image for the Hydra apps. This PR contains changes which make the `hyrax_site.yml` and `sufia_site.yml` playbooks work properly with a Ubuntu 16.04 system.

The most prominent change is the use of the Python3 interpreter for Ansible, which was tripping up `vagrant up`. This change subsequently resulted in the replacement of certain Python2 modules with their Python3 counterparts. Other changes include replacing the upstart init script for `sidekiq` with a systemd init script (taken from [here](https://github.com/mperham/sidekiq/blob/master/examples/systemd/sidekiq.service) and updated for our environment), starting/stopping ClamAV for the database update, and updating the FFmpeg repository URL to Xenial.

I have tested this script using Vagrant and VirtualBox for VTechData, IAWA and Compel. One thing that needs to be tested (for IAWA) is whether Sidekiq is working as intended. The Sidekiq process(es) is up and running, but I don't know whether its interfacing with the Rails app properly.

**Note:** You need to have a version of Ansible `>2.5.0` to use the python3 interpreter properly. The easiest way to do this is by using the `ansible_local` provisioner, and not worry about the version of Ansible or python installed on your host system. To successfully do this, you need to go to the end of `ansible/site_secrets.yml` and make sure `ansible_user: 'vagrant'` is present and uncommented, and then run `ANSIBLE_PROVISIONER=ansible_local vagrant up`.

If you want to use the `ansible` provisioner, you need to upgrade the version of Ansible installed on your system (and probably also have `python3` installed locally for `local_actions`. If you have installed using pip, the command to upgrade Ansible is sudo `pip install ansible --upgrade`. If you have used brew, run `brew update && brew upgrade ansible`.


ansible/roles/common/tasks/main.yml
```diff
@@ -9,13 +9,15 @@
...
-    - python-pip
+    - python3-apt
+    - python3-pip
 
 - name: update python's crypto libs
   pip:
     name: '{{ item }}'
+    state: latest
   with_items:
+    - pip
```
soumikgh, line 20: pip itself needed to be updated or the install of `urllib3` was failing.

soumikgh:
PR 163 takes care of this.

PR #159, opened 2018-05-09
Pin Passenger version to 5.2.x

pmather:
The recently-released 5.3.0 version of Passenger appears to break running Rails applications for our setup.  This commit pins Passenger to version 5.2.x for now.


shabububu (approved):
Approved. I was able to run InstallScripts on the hyrax2 project and view the application from chrome without encountering the passenger error.

PR #160, opened 2018-05-09
Add local actions for "hyrax2" project

pmather:
This adds local actions to be performed when deploying the "hyrax2" project application.

Currently, it adds default roles; adds default users from a list in a file; and upgrades users specified in a file to be admin users.


shabububu (approved):
Approved. I was able to run the install scripts on a hyrax2 branch. The files user_list.txt and admin_lists.txt were copied over to the vm and the rake task was run to add admin roles and populate users. I was able to successfully bring up the application and log in as an administrator.

PR #161, opened 2018-05-10
Fix Ansible 2.5 deprecations

pmather:
In Ansible 2.5 , using Jinja tests as filters is deprecated (https://docs.ansible.com/ansible/2.5/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters).  This leads to a lot of deprecation warnings in the current Ansible playbooks.

This commit fixes those deprecations, leading to "cleaner" playbook runs.


yinlinchen (approved):
Don't see any deprecation warnings.  :thumbsup:

PR #162, opened 2018-05-15
change "is successful" to "is succeeded" in when statements for hyrax2 tasks

whunter:
From this page:
http://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement

### DIFF
--- a/ansible/roles/hyrax/tasks/hyrax2.yml
+++ b/ansible/roles/hyrax/tasks/hyrax2.yml
@@ -21,7 +21,7 @@
     command: bundle exec rails vt_hyrax2:populate_users
     args:
       chdir: '{{ project_app_root }}'
-    when: copy_user_list is successful
+    when: copy_user_list is succeeded
 
   - name: copy list of admins
     copy:
@@ -39,7 +39,7 @@
     command: bundle exec rails vt_hyrax2:upgrade_users
     args:
       chdir: '{{ project_app_root }}'
-    when: copy_admin_list is successful
+    when: copy_admin_list is succeeded
   become: yes
   become_user: '{{ project_runner }}'
   environment:


pmather (commented):
What is wrong with using "successful?"  See recommendation to use "is successful" in the [Ansible 2.5 porting guide](https://docs.ansible.com/ansible/2.5/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters)

whunter:
@pmather I was getting this error before I made the change. The change resolved it.
```TASK [hyrax : add users] *******************************************************
fatal: [hydravm]: FAILED! => {"failed": true, "msg": "The conditional check 'copy_user_list is successful' failed. The error was: template error while templating string: no test named 'successful'. String: {% if copy_user_list is successful %} True {% else %} False {% endif %}\n\nThe error appears to have been in '/Users/leehunter/dev/hyrax2/InstallScripts/ansible/roles/hyrax/tasks/hyrax2.yml': line 20, column 5, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n  - name: add users\n    ^ here\n"}```

**Comment by pmather** [comment]
I am not able to reproduce this error.  What version of Ansible are you using?  You must be using Ansible 2.5 now.

**Comment by whunter** [comment]
```
ansible 2.5.2
  config file = None
  configured module search path = [u'/Users/whunter/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.5.2/libexec/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.15 (default, May 10 2018, 10:34:29) [GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.1)]```

yinlinchen:
@pmather "successful" is deprecated. See [Ansible 2.5 porting guide](https://docs.ansible.com/ansible/2.5/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters) page and scroll up a bit.

pmather:
I see nothing in the porting guide that indicates "successful" is deprecated.  In fact, it states, "In addition to the deprecation warnings, many new tests have been introduced that are aliases of the old tests. These new tests make more sense grammatically with the jinja test syntax, such as the new successful test which aliases success."  (Note that "deprecation warnings" in that context refers to this earlier sentence: "As of Ansible 2.5, using an Ansible provided jinja test with filter syntax, will display a deprecation error.")

Why would they deprecate "successful" in Ansible 2.5 if it was newly-introduced in 2.5?

Note that the "Deprecated" when you scroll up a bit refers to "Jinja tests used as filters."  [Ansible filters](http://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html) use the "|" operator and are used for such things as supplying a default value when a variable is undefined; changing the format of data; or otherwise operating on data.  Previously, "operating on data" also included performing tests, but this is now deprecated in 2.5 in favour of using the "is" operator, which makes the testing aspect more explicit.  That's why in the recent pull request to eliminate deprecation warnings, many (but not all) uses of "|" were changed to "is".

The reason I haven't approved this pull request is that I am unable to replicate the problem Lee is having that led to him submitting it.  Furthermore, the fact that nobody else has encountered this problem leads me to suspect there may be something in Lee's environment that is triggering the issue.  It would be nice to fix that rather than paper over the cracks by merging this pull request.

yinlinchen:
@pmather When I click that page you point, on the left menu. 
screen shot 2018-05-16 at 9 29 49 am

And in @whunter  http://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#the-when-statement, I can't find "successful" anywhere.

I agree the error should be able to be reproduced. Thanks for verifying it.

pmather:
@yinlinchen please read the section you highlight.  Also, please understand what is meant by "Jinja tests used as filters."  (I explain this in my previous comment.)

Also, it is not surprising you don't find "successful" in the documentation on conditionals because this alias is newly introduced in Ansible 2.5 and Ansible documentation is notoriously lagging when it comes to recent releases.  I state this from personal experience as someone who has unfortunately had to use Ansible's documentation.  (On a positive note, it is improving somewhat, so the trend is in the right direction.) 😃

yinlinchen:
@pmather Right, it could be too new and not yet added to their latest document. We should find the root cause and fix it.

PR #163, opened 2018-05-15
Add support for Ubuntu Xenial 16.04

pmather:
This change adds support for deploying to Ubuntu Xenial 16.04 as well as continuing support for deployment to Ubuntu Trusty 14.04. The implementation lays the groundwork for supporting the recently-released Ubuntu Bionic 18.04 LTS in like fashion.

The main approach to supporting multiple OSes is to parameterise settings into different OS-related files and then use "include_vars" in conjunction with "with_first_found" to load the most appropriate settings based upon "ansible_distribution" and other "ansible_distribution_*" values. The order of "with_first_found" is such that more specific OS releases can override less general ones, with "defaults.yml" being a catch-all for settings common to all supported releases.

In Vagrant terms, the box used will determine the "ansible_distribution_*" values.  It currently defaults to "bento/ubuntu14.04" but can be overridden by setting the VAGRANT_VM_BOX environment variable.

For a limited set of boxes, the Vagrantfile knows how to map the box name to the corresponding Ubuntu release.  This is needed because Ubuntu Xenial comes with Python 3 installed.  As such, "ansible_python_interpreter" must be added as an extra variable in the inventory for the guest VM host being deployed to in the case of a Xenial Vagrant box.  Vagrant's "host_vars" is used to accomplish this purpose.


soumikgh:
I have looked at the changes, and deployed apps using the `sufia_site`, `hyrax_site`, `geoblacklight_site` and also a `hyrax2` app. Everything seems to be working fine, and the code changes look good. I think this is good to be merged, despite the issues with `resque-pool`, because those seem to have been present even without the changes from this PR.

soumikgh (approved):
I have looked at the changes, and deployed apps using the `sufia_site`, `hyrax_site`, `geoblacklight_site` and also a `hyrax2` app. Everything seems to be working fine, and the code changes look good. I think this is good to be merged, despite the issues with `resque-pool`, because those seem to have been present even without the changes from this PR.

PR #166, opened 2018-06-12
add rake tasks for compel application

whunter:


ansible/roles/hyrax/tasks/compel.yml
```diff
@@ -0,0 +1,44 @@
+---
+- block:
+  - name: load workflow
+    command: ./bin/rails hyrax:workflow:load
```
pmather: This Rake task is executed for all Hyrax applications as part of the `hyrax/tasks/main.yml` play sequence in the `hyrax` role.  It is superfluous here.

ansible/roles/hyrax/tasks/compel.yml
```diff
@@ -0,0 +1,44 @@
...
+    args:
+      chdir: '{{ project_app_root }}'
+
+  - name: create default collection type
+    command: ./bin/rails compel:create_default_collection_type
```
pmather: Stylistically, `bundle exec` is preferred over `./bin/`.  E.g., `bundle exec rails ...` rather than `./bin/rails ...`.

ansible/roles/hyrax/tasks/compel.yml
```diff
@@ -0,0 +1,44 @@
...
+    notify: clear rails cache
+
+  - name: add new font files
+    unarchive:
+      src: '{{ local_files_dir }}/fonts.zip'
+      dest: '{{ project_app_root }}/app/assets/fonts/'
+    ignore_errors: True
+    notify: clear rails cache
+  become: yes
+  become_user: '{{ project_owner }}'
```
pmather, line 42: Rake tasks affecting internal application state should usually be executed as `project_runner` whereas file system operations (like copying files, making directories, and unpacking .zip files) should be executed as `project_owner` so as to better observe permissions.  See, e.g., the `hyrax/tasks/iawa.yml` local tasks plays for an example of this.  (It uses different users across different blocks.)

pmather (changes requested):
This looks like it is based somewhat off the `binary-arts.yml` local tasks.  Things have changed a little in the `hyrax` role since then, hence the slight changes requested.

whunter:
@pmather Changes made

pmather (approved):
👍

PR #167, opened 2018-06-12
LIBTD-1348: Fix permission issue when branding a collection

tingtingjh:
branding image permission

branding logo


pmather (approved):
If branding is a common feature (which I think it is from what you said recently) then it's good to add this here.  If it is a feature you've added to a specific application code base (i.e., it doesn't come as part of Hyrax) then I would say it belongs better in an application's local tasks.

tingtingjh:
@pmather Yes, I believe it is a common feature.

shabububu (approved):
I approve. I tested it out with the compel project, and the permissions error did not happen when I attempted to brand a collection.

screen shot 2018-06-12 at 11 17 25 pm

PR #169, opened 2018-07-02
change ezid to doi and add "host" config

whunter:


pmather (approved):
I tested this with the `hyrax2` application and the `doi` elements were populated correctly in `config/secrets.yml`.

Note, the `Rails.application.secrets['ezid']` references in `VTUL/data-repo` need to be changed to reference `doi` instead.  Currently, the `ezid` references cause the `db_migrate` Rake task to fail during deployment due to the attempts to derefernce keys from a NIL dictionary.

whunter:
@pmather I have a branch in data-repo that references doi instead of ezid in the secrets hash. Will start testing it now that this is merged. Thank you!

PR #170, opened 2018-07-03
fonts for compel are tracked in git repo. Don't delete them on deploy

whunter:
Compel fonts are non-proprietary so we can just keep them in the git repo.

https://fonts.google.com/specimen/Raleway


pmather (approved):
Tested; skips font copying/deletion.

👍

PR #171, opened 2018-07-11
Sunset EZID settings

pmather:
The previous DOI provider, EZID, had config/secrets.yml referenced under the key space of "ezid".  When this was changed to the more generic key space of "doi", older applications that still referenced "ezid" broke due to the absence of that key hierarchy in config/secrets.yml.

This change restores the old "ezid" settings alongside the new, preferred "doi" settings, for backwards compatibility with applications that have yet to update to using "doi".

Note that the "ezid" keys are now deprecated and should be removed in the future.

### DIFF
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -53,6 +53,12 @@ graylog_verbosity: "info"
 #
 # The GeoNames user the application will use (if any)
 project_geonames_user: ''
+# EZID shoulder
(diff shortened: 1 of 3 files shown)


shabububu:
I was able to build the iawa master application successfully with these changes. I'm not sure how to test out the DOI related functionality, or if it's even necessary for this PR.

shabububu (approved):

PR #172, opened 2018-07-16
LIBTD-1418: Short-term fix: Compel Default CollectionTypes

shabububu:
To be tested with https://github.com/VTUL/compel/tree/LIBTD-1418

Temporary solution in order to create the default Hyrax Collection Types.

One way of testing it out is to:
1. Sign up as a user
2. Create an `admin_list.txt` file with your email address:
   ```
   echo "me@example.org" > admin_list.txt
   ``` 
3. Run the rake task to elevate your user to admin status:
   ```
   bin/rails compel:upgrade_users
   ```
4. Check out the Collection Types in the dashboard to see if "User Collection" and 
Admin Set" have been created properly.

screen shot 2018-07-16 at 4 21 18 pm

### DIFF
--- a/ansible/roles/hyrax/tasks/compel.yml
+++ b/ansible/roles/hyrax/tasks/compel.yml
@@ -1,7 +1,7 @@
 ---
 - block:
-  - name: create default collection type
-    command: bundle exec rails compel:create_default_collection_type
+  - name: create default hyrax collection type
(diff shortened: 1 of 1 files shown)


pmather (approved):
I tested this in `RAILS_ENV=production` mode against the `LIBTD-1418` branch of `VTUL/compel` and got the same collection output reported for me.

PR #173, opened 2018-07-23
LIBTD-1449 - Add pull request template

yinlinchen:
Add a PR template for VTUL projects and use this repository as a start.


shabububu:
Would this the right place to also include other useful information like when to squash commits/comments? If not, where is a good place to put that information? I think it's useful to have conventions documented somewhere so that relatively new folks like me know what normal protocol is.

yinlinchen:
@shabububu You were talking about git workflow, we can create a git workflow document in the confluence.

shabububu:
Are some of the sections optional? Or, are all of them required? That might be useful to note.

pmather (approved):
The consensus in offline discussion is that this is good to go.

PR #177, opened 2018-09-18
Add support for installing Google Chrome

pmather:
**Add support for installing Google Chrome**
* * *

Add a role to enable Google Chrome to be installed.  This is used in the Rails "test" and "development" environments to run feature tests using Google Chrome in "headless" mode via "chromedriver".

# What's the changes?

* Add `chrome` Ansible role that may be used to install Google Chrome
* Amend `hyrax` role to add `chrome` role as a dependency (so it will be installed) except when deploying in `production` mode

# How should this be tested?

* Set `ansible/site_secrets.yml` to deploy a Hyrax application, e.g., `compel`
* Run `vagrant up` to deploy the application
* Log into the resultant local VM via `vagrant ssh`
* Run `which google-chrome` command
* Run `google-chrome --version` command

# Additional Notes:

The expected output from the last two commands above is as follows:
```
vagrant@compel:~$ which google-chrome
/usr/bin/google-chrome
vagrant@compel:~$ google-chrome --version
Google Chrome 69.0.3497.100
```

# Interested parties

@shabububu

### DIFF
--- a/ansible/roles/chrome/README.md
+++ b/ansible/roles/chrome/README.md
@@ -1,0 +1,14 @@
+Chrome
+======
+
+Installs the Google Chrome browser.
+
+Requirements
+------------
+
+This requires our Common role.
+
+Role Variables
+--------------
+
+- `chrome_branch`: The desired release stream of Google Chrome: `stable`, `unstable`, or `beta`.  (Default: `stable`)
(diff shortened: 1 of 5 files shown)


shabububu (approved):
I successfully used InstallScripts to build the compel project in both development and test.  Both environments had chrome installed with the matching version.

PR #179, opened 2018-10-18
Fix Ansible 2.7 apt/pip deprecations

pmather:
**Fix Ansible 2.7 apt/pip deprecations**
* * *

Ansible 2.7 now deprecates the use of loops to install multiple packages with "apt" or "pip".  Instead, the preferred method is to supply a list of packages when multiple packages are being installed.

This commit changes all uses of loops ("with_items" constructs) in conjunction with the "apt" or "pip" modules to use lists or list variables instead.  The deprecated loops feature will be removed entirely in Ansible 2.11.

# What's the changes?

All current use of the "with_items" looping mechanism has been removed in the case of the "apt" and "pip" modules and replaced with lists instead. This silences deprecation warnings relating to that idiom.

# How should this be tested?

Deploy an application using InstallScripts using both Ansible 2.6 and 2.7.  There should be no `[DEPRECATION WARNING]` log entries and the application should deploy successfully in either case.

# Interested parties
@soumikgh

### DIFF
--- a/ansible/roles/clamav/tasks/main.yml
+++ b/ansible/roles/clamav/tasks/main.yml
@@ -1,13 +1,10 @@
 ---
 - name: install clamav
   apt:
-    name: '{{ item }}'
+    name: ['clamav', 'libclamav-dev']
     state: present
     cache_valid_time: '{{ apt_cache_timeout }}'
     update_cache: yes
-  with_items:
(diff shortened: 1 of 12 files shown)


soumikgh:
The changes look good and I tested by deploying Compel.

PR #181, opened 2018-11-07
remove protocol and trailing slash. update shoulder

whunter:
**Update datacite defaults in example site secrets.**
* * *

**JIRA Ticket**: (https://webapps.es.vt.edu/jira/browse/LIBTD-1620) (:star:)

# What does this Pull Request do? (:star:)
Update datacite defaults in `ansible/example_site_secrets.yml` to correct errors

# What's the changes? (:star:)
* remove protocol and trailing slash from project_doi_host.
* update doi shoulder

# How should this be tested?
* Provision a VM using the master branch of `data-repo` using Datacite user/pass that can be provided
* Run rspec tests and check that there are no doi related failures

# Additional Notes:
* branch: `datacite_defaults`

# Interested parties
@pmather 

(:star:) Required fields

### DIFF
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -60,9 +60,9 @@ project_ezid_user: 'apitest'
 # EZID API password
 project_ezid_password: 'apitest'
 # DOI host
-project_doi_host: 'https://ez.datacite.org/'
+project_doi_host: 'ez.datacite.org'
 # DOI shoulder
-project_doi_shoulder: 'doi:10.5072/FK2'
+project_doi_shoulder: 'doi:10.7294/W4'
 # DOI API username
 project_doi_user: 'apitest'
 # DOI API password


pmather (commented):
I have tested this change but I get DOI-related test failures when following the "How should this be tested?" instructions.  However, I don't believe the failures are related to this change, necessarily, but to the tests themselves, which appear to test against hard-coded DOIs that use the old shoulder (e.g., `doi:10.5072/FK2AB1C234`).  This relates to the `:my_collection_with_doi` definition in `spec/models/ability_spec.rb`.

I'm okay with approving this change, because it, in itself, does not appear the source of the problem.  That lies with the test in `VTUL/data-repo`, which, I see, seems to have been changed (albeit hard-coded to the new value) in the `dev` branch, which doesn't appear to have made it into the `master` branch yet.

(It could be considered an error to hard-code the DOI string in the test when its prefix is derived from a `Rails.application.secrets` value.)

pmather (approved):
Following a discussion at today's meeting, I am approving these changes because, as already mentioned, the test failures are not related strictly to this change but to the test itself.  It was suggested that the test be rewritten in the future to address these concerns.

PR #182, opened 2018-11-26
Add support for Ubuntu 18 (Bionic)

pmather:
**Add support for Ubuntu 18 (Bionic)**
* * *

This change adds support for deploying applications under Ubuntu 18 (Bionic).

# What's the changes?
There are several important changes in Bionic that necessitated some refactoring of the existing roles beyond just adding some Bionic-specific variables:

- Tomcat 7 is not available as a base OS package: Tomcat 8 is available instead. This caused the `tomcat7` role to be refactored into a more generic `tomcat` role that supports both Tomcat 7 (for Trusty and Xenial) and 8 (for Bionic).

- The way Nginx and Passenger are installed under Bionic differs to the approach used previously on Trusty and Xenial. Under Bionic, the modularised version of Nginx is used, not the version with statically-included modules. This required a refactoring of the `passenger` role.

Other than these major changes, Bionic support was accommodated in the way that Xenial support was previously, i.e., via OS-version specific variables files.

# How should this be tested?

The easiest way to test this is to deploy applications via `VTUL/InstallScripts` using Vagrant. You should set `VAGRANT_VM_BOX` accordingly in your environment prior to `vagrant up` so as to select an Ubuntu 18.04 Bionic base box and then re-test by setting it to an Ubuntu 16.04 Xenial base box. The deployed application should work under both versions of the underlying OS.

Example:
1. Edit `ansible/site_secrets.yml` to choose an appropriate application to deploy.
2. `export VAGRANT_VM_BOX=bento/ubuntu-18.04`
3. `vagrant up`
4. Make sure application works.
5. `vagrant destroy --force`
6. `export VAGRANT_VM_BOX=bento/ubuntu-16.04`
7. Repeat steps 3 to 5.

# Additional Notes:
Deploy branches of applications known to work currently under Ubuntu 16.04.

These changes try to diverge as little as possible between the different OSes supported.

# Interested parties
@soumikgh


soumikgh (approved):
I have deployed IAWA, Compel, GeoBlacklight, and VTechData with both Ubuntu 16.04 and 16.04 and the apps get successfully deployed and come up when requested from the browser.

I had a brief look at the code changes and they look good too!

PR #183, opened 2018-11-30
Support locally-supplied Fedora WAR file

pmather:
**Support locally-supplied Fedora WAR file**
* * *

Allows a locally-supplied Fedora WAR file to be used when deploying Fedora 4.

# What's the changes?
Up to now, only official distribution Fedora WAR files can be used when installing Fedora 4.  In some cases, it is desirable to use a locally-patched or otherwise locally-produced Fedora WAR file instead.

This change allows a Fedora WAR file to be placed in the `local_files` directory and used in preference to the distribution version when installing Fedora 4.  If a local WAR file is not present, it will fall back on using the distribution version, as normal.

The `fedora_war_filename` Ansible variable denotes the name of the Fedora WAR file that is used, if present, in `local_files`.  This filename is also used for the fall-back distribution WAR file, so the `local_files` version should have the same name as the Fedora distribution version.

# How should this be tested?

Place a copy of the `fcrepo-webapp-4.5.1.war` file in the `local_files` directory and deploy a Samvera application.  The application should deploy using the locally-supplied WAR file.

Repeat the process afresh, this time without the local WAR file present.  The application should deploy using the fall-back distribution version.

# Interested parties
@soumikgh

### DIFF
--- a/ansible/roles/fedora4/README.md
+++ b/ansible/roles/fedora4/README.md
@@ -18,6 +18,8 @@ Role variables are listed below, along with their defaults:
     fedora_user_home: /var/local/<tomcat_user>
     fedora_data_dir: /var/local/<tomcat_user>/fedora-data
     fedora_app_dir: /var/lib/<tomcat_user>/webapps
+    fedora_war_filename: fcrepo-webapp-<fedora_version>.war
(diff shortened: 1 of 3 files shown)


soumikgh:
The code works as intended and uses the locally-supplied war file if its there, otherwise fetches it from a remote location.

PR #189, opened 2019-01-28
Default to Ubuntu 18.04 for Vagrant

pmather:
**Default to Ubuntu 18.04 for Vagrant**
* * *

InstallScripts supports deploying to three versions of Ubuntu: 14.04, 16.04, and 18.04.  When deploying via Vagrant, the version to use can be set explicitly by setting the `VAGRANT_VM_BOX` environment variable to a supported Vagrant box prior to executing `vagrant up`.  If `VAGRANT_VM_BOX` is not set then a default is used.

This change switches the default to an Ubuntu 18.04 Vagrant box from the current Ubuntu 16.04 box that is used.

# What's the changes?

Henceforth, `vagrant up` will use the `bento/ubuntu-18.04` box by default.

# How should this be tested?

Deploy an InstallScripts application (e.g., `geoblacklight`) using Vagrant.  The application should deploy successfully into an Ubuntu 18.04 local VM.  You can verify the local VM is running Ubuntu 18.04 as follows after a successful `vagrant up` has completed:

1. `vagrant ssh`
2. `lsb_release -a`

The `lsb_release -a` command should return information indicating the VM OS is Ubuntu 18.04:
```
vagrant@geoblacklight:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
vagrant@geoblacklight:~$
```

# Additional Notes:
For completeness, after testing the default Ubuntu 18.04 deployment, you can `vagrant destroy --force` the VM and then deploy the same application in an Ubuntu 16.04 VM:

1. `vagrant destroy --force`
2. `export VAGRANT_VM_BOX=bento/ubuntu-16.04`
3. `vagrant up`
4. `vagrant ssh`
5. `lsb_release -a`

The output of `lsb_release -a` should indicate the VM OS is Ubuntu 16.04:
```
vagrant@geoblacklight:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.5 LTS
Release:	16.04
Codename:	xenial
vagrant@geoblacklight:~$
```
# Interested parties
@VTUL/dld-dev

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -13,7 +13,7 @@
 # If no "--provider" is specified during "vagrant up" then the default
 # (VirtualBox) provider will be used.
 
-$vm_box = ENV['VAGRANT_VM_BOX'] || 'bento/ubuntu-16.04'
+$vm_box = ENV['VAGRANT_VM_BOX'] || 'bento/ubuntu-18.04'
 $secrets_items = {}
 begin
   require 'yaml'


shabububu (commented):
To try this out with IAWA, I checked out the `default_to_bionic` branch and then did a `git merge master` before creating a local production with the `iawa` project.

Using `vagrant ssh`, I checked the VM OS:

```
vagrant@iawa:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

In terms of application functionality, I tried:
* Manually adding a collection and work via the UI.
* Performed a basic batch import via the UI.
* Viewed the works and collections.

For the application, everything worked as expected. 

One thing I noticed is that when I initially tried to change users, I lost the bash shell. This is one difference compared to ubuntu 16 in terms of how I've been using it.

```
vagrant@iawa:~$ sudo su - hydra
$ bash
hydra@iawa:~$
```

I also ran the steps in the "Additional Notes", and the result was as expected.

Finally, I tried a local `development` build, and repeated the steps from production (without the Additional Notes steps). Everything was as expected. Also, ran `bundle exec rspec` and the tests seemed to run fine.

Note: I'm not going to merge because I only tested iawa.

pmather:
In investigating the shell behaviour you mention above, I believe there is a bug in the way the Ansible `user` module (or the underlying tools it invokes) is working on Ubuntu 16.04---at least it does not appear to follow the documentation.

The Ansible `user` module employs `useradd` on Ubuntu to do the job of actually adding the user.  The `useradd` manual page has this to say about the shell for the user:
```
       -s, --shell SHELL
           The name of the user's login shell. The default is to leave this field blank, which causes the system to select the
           default login shell specified by the SHELL variable in /etc/default/useradd, or an empty string by default.
```
The `/etc/default/useradd` file on both Ubuntu 16.04 and 18.04 have `SHELL=/bin/sh`.  However, the actual user created on Ubuntu 16.04 has an empty shell field in the password file.  On Ubuntu 18.04 it correctly has the default `/bin/sh` for the shell.

I don't know what is causing Ubuntu 16.04 to fall back on using `/bin/bash` in the absence of a shell in the password file, but that is likely what is causing `/bin/bash` to be used on Ubuntu 16.04 and `/bin/sh` (explicitly specified as the user's shell in the password file) on Ubuntu 18.04.

PR #191, opened 2019-02-08
Wire down uid/gid for users

pmather:
**Wire down uid/gid for users**
* * *

InstallScripts adds several users to the system during deployment: two for Rails applications and one as an SFTP uploader (for `geoblacklight`).  Currently, only the user and group names are specified when creating these users.  The actual uid and gid used is auto-assigned by the system so as not to conflict with existing users.  If users have been created on the OS outside InstallScripts then this makes the actual uid and gid settings of the InstallScripts users unpredictable.  This, in turn makes it difficult to align uid and gid assignment for, e.g., NFS mounts or other situations where uid and gid settings must match precisely.

This change allows uid and gid to be specified precisely for InstallScripts-added users.  (Note, this does not apply to users created by installed packages, e.g., Solr, Tomcat, etc., which manage their own associated users.)  The defaults for uid and gid start at 60000, which lies outside the range of system auto-assigned uid and gid, so as not to conflict.

The settings in `example_site_secrets.yml` have been updated to values that will work correctly for all currently-supported InstallScripts applications.

# What's the changes?

* Adds variables to specify uid and gid settings for `project_owner`, `project_runner`, and `sftp_user` users created and managed by InstallScripts
* Updates `example_site_secrets.yml` with working values for uid and gid

# How should this be tested?

* Deploy an InstallScripts-supported application
* Check it works

# Additional Notes:

* Testing the `geoblacklight` application will test all InstallScripts-managed users
* Ansible Tower templates will have to be updated to add these extra settings

### DIFF
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -18,14 +18,23 @@ project_secret_key_base: ''
 #
 # Name of the user who owns/edits the application
 project_owner: 'hydra'
+# The numeric uid of {{ project_owner }}
+project_owner_uid: 60001
 # Name of group to which {{ project_owner }} belongs
 project_owner_group: '{{ project_owner }}'
+# The numeric gid of {{ project_owner_group }}
+project_owner_gid: 60001
 # The home directory of {{ project_owner }}
 project_owner_home: '/var/local/{{ project_owner }}'
 # Name of the user running the application (should be different from project_owner)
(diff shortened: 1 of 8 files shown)


soumikgh:
I can confirm the the UID/GID gets assigned correctly for the `project_owner` and `project_runner` users. Merging.

PR #193, opened 2019-04-16
Add support for smart relay host

pmather:
**Add support for smart relay host**
* * *

Amends the `postfix` role to allow a smart relay host to be defined. The variable `mail_relayhost`, if defined, is used to define the Postfix `relayhost` parameter. If `mail_relayhost` is not defined then any
definition of `relayhost` is removed from the Postfix configuration (and Postfix sends e-mail directly instead of via a smart host).

# What's the changes?
Sets or removes the Postfix `relayhost` setting from its `main.cf` file depending upon whether
`mail_relayhost` is defined.

# How should this be tested?

1. Deploy the `VTUL/data-repo` application using this branch.  A `grep relayhost /etc/postfix/main.cf` inside the resultant VM after `vagrant up` should not return any results.
2. Next, do a `vagrant provision` of the application but with the `mail_relayhost: mailrelay.smtp.vt.edu` line from `site_secrets.yml` uncommented this time. Now,  `grep relayhost /etc/postfix/main.cf` should return a line with `relayhost = [mailrelay.smtp.vt.edu]`.

# Interested parties
@soumikgh

### DIFF
--- a/ansible/example_site_secrets.yml
+++ b/ansible/example_site_secrets.yml
@@ -174,7 +174,9 @@ sftp_upload_root: "/opt/sftp/geodata"
 # Fetch local_files from remote URL
 # Remote archive URL - remote archive fetched only if this variable is set
 # local_files_archive: 'https://hydrafiles.lib.vt.edu/{{ project_name }}/local_files.zip'
-#
+# Uncomment and assign appropriately the setting below to define a smart mail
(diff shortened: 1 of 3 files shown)


soumikgh:
I can confirm that a relayhost is set when the `mail_relayhost` variable is defined in `site_secrets.yml`.

PR #195, opened 2019-06-04
Improve switching of Ruby version

pmather:
**Improve switching of Ruby version**
* * *

The repository from which we install Ruby supports multiple versions. We
allow the version to be selected by setting the `ruby_version` variable.
Unfortunately, changing the value of `ruby_version` does not mean the
new version of Ruby is used. This change fixes that behaviour, ensuring
explicitly that the version of Ruby selected will be used by the relevant
running Ruby applications.

# What's the changes?

The `ruby-switch` utility is used with the packaged versions of Ruby to
determine which version of Ruby installed will be the default system
Ruby. Previously, whichever version of Ruby was installed first would
become the version of Ruby used from there on. Using `ruby-switch`
after installing (or changing) Ruby versions ensures the latest
installed version will be used as default.

In addition, when the version of Ruby is switched, we need explicitly to
restart any running Ruby-based daemons/services. This change introduces
an explicit test of whether the current version of Ruby being deployed
changes from that which was already deployed. If this is the case, a new
fact is set that may be tested in other Ruby-consuming roles to know
whether or not to restart long-running daemons or services that use Ruby
installed by these playbooks.

# How should this be tested?

Deploy a supported application. This will use the default Ruby version 2.3. Then, edit `site_secrets.yml` and append the line `ruby_version: "2.4"`. Redeploy the application (e.g., `vagrant provision`). This will change the version of Ruby to 2.4.

# Additional Notes:

You can use `ruby --version` to identify the default version of Ruby that is used.  This should be 2.3.x after the first deployment and 2.4.x after the second.

# Interested parties
@soumikgh

### DIFF
--- a/ansible/roles/geoblacklight/tasks/main.yml
+++ b/ansible/roles/geoblacklight/tasks/main.yml
@@ -13,3 +13,8 @@
 - include: geoblacklight_setup.yml
 - include: solr_setup.yml
 - include: cron.yml
+
+- name: restart ruby applications when ruby version changed
+  command: /bin/true
+  notify: restart nginx
+  when: ruby_version_changed | default( False )
--- a/ansible/roles/hyrax/tasks/main.yml
+++ b/ansible/roles/hyrax/tasks/main.yml
@@ -158,6 +158,9 @@
     clean: yes
     exclude_groups: ''
   when: project_app_env != 'production'
+  notify:
+    - restart nginx
+    - restart sidekiq workers
   become: yes
   become_user: '{{ project_owner }}'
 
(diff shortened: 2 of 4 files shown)


soumikgh:
I can confirm that adding `ruby_version: "2.4"` to the site_secrets file leads to a switch in the default Ruby version set in the VM.

```
vagrant@ubuntu-bionic:~$ ruby --version
ruby 2.3.8p459 (2018-10-18 revision 65136) [x86_64-linux-gnu]

# After re-provision
vagrant@ubuntu-bionic:~$ ruby --version
ruby 2.4.6p354 (2019-04-01 revision 67394) [x86_64-linux-gnu]```

PR #199, opened 2019-09-12
make tmp/cache writable and delete bootsnap cache to fix error

whunter:
**Make tmp/cache writable and delete bootsnap cache to fix error.**
* * *

**JIRA Ticket**: (https://webapps.es.vt.edu/jira/browse/LIBTD-1890) (:star:)

# What does this Pull Request do? (:star:)
Newest version of Geoblacklight leverages the bootsnap gem (https://github.com/Shopify/bootsnap) to speed up the application. This gem needs to be able to write to the application's `tmp/cache` directory. Bootsnap also throws an error that prevents the application from starting up. Deleting it's cache and letting it be rebuilt as the application is used resolves this.

# What's the changes? (:star:)
* Makes `(application_root)/tmp/cache` writable
* Deletes bootsnap cache before initial startup

# How should this be tested?
* Build Geoblacklight application using `geoblacklight_update` branch of Installscripts and `updated` branch of vtul/geoblacklight
* Ensure that the application will load in browser. (More thorough application testing can be handled in the vtul/geoblacklight update PR to come)

# Additional Notes:
* Branches: 
  * Installscripts: `geoblacklight_update`
  * VTUL/geoblacklight: `updated`

# Interested parties
@pmather 

(:star:) Required fields

### DIFF
--- a/ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
+++ b/ansible/roles/geoblacklight/tasks/geoblacklight_setup.yml
@@ -38,6 +38,8 @@
   with_items:
     - path: "{{ project_app_root }}/tmp"
       mode: "0777"
+    - path: "{{ project_app_root }}/tmp/cache"
+      mode: "0777"
     - path: "{{ project_app_root }}/log"
       mode: "0770"
     - path: "{{ project_app_root }}/db"
--- a/ansible/roles/geoblacklight/tasks/main.yml
+++ b/ansible/roles/geoblacklight/tasks/main.yml
@@ -18,3 +18,11 @@
   command: /bin/true
   notify: restart nginx
   when: ruby_version_changed | default( False )
+
+- name: remove Bootsnap load path cache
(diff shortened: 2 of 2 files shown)


pmather (approved):
I tested this branch against the `updated` branch of `VTUL/geoblacklight`.  The resultant application runs in the browser.

Note, the `Gemfile` in the `updated` branch of `VTUL/geoblacklight` contains a Ruby version check against version 2.5.5.  This requires specifying `ruby_version: 2.5` in `site_secrets.yml` when deploying (otherwise it fails to deploy with a `Your Ruby version is 2.3.8, but your Gemfile specified 2.5.5` error).  I would recommend that the `ruby '2.5.5'` line in the `updated` branch `Gemfile` be amended to `ruby '~> 2.5.5'` to avoid a strict dependency on Ruby 2.5.5.  I'm presuming that GeoBlacklight needs just any version of Ruby 2.5 to run?  Maybe also it is time to change the `ruby_version` default in `example_site_secrets.yml` to `2.5`?

